PR #4, opened 2025-06-12
Setting up Detekt Ktlint wrapper

hamorillo:
### Description

Adding [Detekt Ktlint wrapper](https://detekt.dev/docs/rules/formatting/) to verify the code format in the project.

### Testing Steps

- Check out this branch and execute:

```
gradlew detetk
```

or 
```
gradlew detekt --auto-correct
```

Verify it works. You can introduce an error in the code and reexecute it to verify that it's working as expected.


wzieba:
👋 WDYT about using Detekt with ktlint as a formatter? This is something we use across other apps. Usually adding Detekt earlier is easier, from introducing it later and fixing issues.

wzieba:
Ah sorry, I see you added Detekt already :) so my suggestion would apply to not introduce ktlint Gradle plugin, but simply the ktlint formatter _into_ Detekt

.editorconfig
```diff
@@ -0,0 +1,11 @@
+# https://editorConfig.org
+root = true
+
+[*.{kt,kts}]
+ktlint_code_style = ktlint_official
+ktlint_function_naming_ignore_when_annotated_with = Composable
+max_line_length=120
```
AdamGrzybkowski: 💡 I would go with 140.

.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+ktlint_function_signature_body_expression_wrapping = default
+ktlint_function_signature_rule_force_multiline_when_parameter_count_greater_or_equal_than = 6
+ktlint_standard_multiline-expression-wrapping = disabled
+ktlint_standard_string-template-indent = disabled
```
AdamGrzybkowski: ⛏️ Missing new line

hamorillo:
> Ah sorry, I see you added Detekt already :) so my suggestion would apply to not introduce ktlint Gradle plugin, but simply the ktlint formatter into Detekt

Thanks for your suggestion @wzieba. I've replaced Ktlint dependency with Detekt's Ktlint wrapper.

.editorconfig
hamorillo: Done!

hamorillo:
Apart from `MaximumLineLength`, I kept the default formatting rules. If you have strong opinions about modifying any of them, please let me know. Anyway, we can always iterate if needed. 😄

AdamGrzybkowski (approved):
RIP my muscle memory to type `gw ktlintFormat` 😅

PR #7, opened 2025-06-16
Introduce GravatarAndroidApplicationConventionPlugin

AdamGrzybkowski:
### Description

This PR introduces the convention plugins in the repo. For now, there's only one: `GravatarAndroidApplicationConventionPlugin`. As we add more modules, we will have to add more for AndroidLibrary and KotlinLibrary at least.

I've also set up detekt for the build-logic module. 

### Testing Steps

1. Build the project
2. Run the app
3. You can verify in the apk manifest that things like minSDK are configured properly
4. Run `./gradlew detekt --auto-correct` to verify the formatting still works


build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
```diff
@@ -0,0 +1,51 @@
...
+                defaultConfig.apply {
+                    applicationId = APP_ID
+                    targetSdk = TARGET_SDK
+                    versionCode = 1
+                    versionName = "1.0"
+                }
+                configureBuildTypes()
+                buildFeatures {
+                    compose = true
+                }
```
AdamGrzybkowski, line 34: I've kept the compose setup here for now - we might want to extract it to a separate plugin once we need it in more places and we use more config options.

AdamGrzybkowski:
hm @wzieba shouldn't the CI build this PR when I forced pushed after rebasing with the trunk?

wzieba:
You're right @AdamGrzybkowski , I'm working on the fix 👍

wzieba:
> Run ./gradlew :build-logic:convention:detekt --auto-correct to verify formatting work in the build-logic module

Would it be possible to introduce custom source to base `detekt` task instead? I think it'd be nice to have one `detekt` task that verifies the whole codebase. Then, I think you even wouldn't have to apply `detekt` to `build-logic` project.

wzieba:
The checks should now work @AdamGrzybkowski . The suggestion I meant in my last comment is this: https://github.com/Automattic/Gravatar-Android/compare/adam/GRA-182...adam/GRA-182-suggestion?expand=1

hamorillo (approved):
Sorry for the delay, it's my first time working with the Convention Plugins [TIL]

Looks good! 🚀

AdamGrzybkowski:
@wzieba Yeah, that's a good idea. I've cherry-picked your commit here.

PR #10, opened 2025-06-17
Basic Root navigation for the app

AdamGrzybkowski:
### Description

This PR adds a basic root navigation for the App using the Jetpack Compose library. The `:app` module serves as a glue for the app component here, meaning it will put together all the Koin modules and setup the root destination. 

There are 3 root destinations for now: splash, login, and home.

I've already put login and home in separate modules and exposed the main composable and the Koin module from it. 
The idea is to use "package by component" architecture, you can read about it [here](https://medium.com/clean-android-dev/the-real-clean-architecture-in-android-modularization-e26940fd0a23). I'm open to discussing whether this is what we want to pursue. 

### Testing Steps

Launch the app and try navigating between the screens by tapping on the buttons.


build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
```diff
@@ -0,0 +1,46 @@
...
+            extensions.configure<LibraryExtension> {
+                configureKotlinAndroid(this)
+                compileSdk = COMPILE_SDK
+                defaultConfig.apply {
+                    consumerProguardFiles("consumer-rules.pro")
+                }
+                configureBuildTypes()
+                buildFeatures {
+                    compose = true
+                }
```
hamorillo: Not all libraries will require compose. Maybe we should have different plugins.

hamorillo (approved):
Looks good. :shipit: 

Thanks for the link to the Medium post.

build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
AdamGrzybkowski: True, I saw that in your PR. I will extract it here before merging.

AdamGrzybkowski: @hamorillo Done, do you want to review again?

hamorillo: Looks good! Feel free to merge.

PR #11, opened 2025-06-17
Create Analytics module

hamorillo:
### Description

This PR creates a new module (`analytics`) to encapsulate the code related to analytic events. [Track library doesn't have a userType for Gravatar](https://github.com/Automattic/Automattic-Tracks-Android/blob/c0193681980308468810a6828882e3742257d3f2/AutomatticTracks/src/main/java/com/automattic/android/tracks/TracksClient.java#L42), so I've opened a [PR](https://github.com/Automattic/Automattic-Tracks-Android/pull/263) to add it. (We'll need to update our app when the PR is merged and the new release is published.)

If we agree to merge this PR, I'll create a ticket to ensure we don't forget about it.

### Testing Steps

- I've added a "test" event in the viewmodel that we'll need to remove when we have a clear idea about what event we want to track. Launching the app and filtering tracks by `gravatarandroid%` should be enough to verify that the event is being sent to the backend. Another option could be to verify that the device makes the network request.

**Note:** The event takes some time to appear in Tracks. Based on my experience, it could take more than 15 minutes.


analytics/src/main/java/com/gravatar/analytics/tracks/TracksTracker.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.gravatar.analytics.Tracker
+import java.util.UUID
+
+internal class TracksTracker(private val tracksClient: TracksClient) : Tracker() {
+    override var userId: String? = null
+    private val anonId: String = generateNewAnonID()
+
+    override fun trackEvent(event: Event) {
+        // We should add GRAVATAR userType when available.
+        tracksClient.track(event.name, userId ?: anonId, TracksClient.NosaraUserType.ANON)
```
hamorillo: We should refactor the code when `NosaraUserType.GRAVATAR` is available.

analytics/src/main/java/com/gravatar/analytics/tracks/TracksTracker.kt
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    override fun flush() {
+        tracksClient.flush()
+    }
+}
+
+private fun generateNewAnonID(): String {
+    // Generate a new UUID and return it as a string.
+    return UUID.randomUUID().toString()
```
hamorillo: We should probably store locally the AnonId always to use the same one in case the user doesn't log in the first time. I would keep this to a following PR.

analytics/src/main/java/com/gravatar/analytics/Event.kt
```diff
@@ -0,0 +1,5 @@
+package com.gravatar.analytics
+
+interface Event {
```
hamorillo: I've created an `interface` instead of a `sealed` class here to let each module declare its events.

app/src/main/java/com/gravatar/app/MainActivityViewModel.kt
```diff
@@ -1,11 +1,17 @@
...
+import com.gravatar.analytics.Tracker
+import com.gravatar.app.analytics.AppEvent
 
-class MainActivityViewModel : ViewModel() {
+class MainActivityViewModel(private val tracker: Tracker) : ViewModel() {
 
     @Suppress("FunctionOnlyReturningConstant")
     fun getName(): String {
-        return "Gravatar"
+        val userId = "Gravatar"
```
hamorillo: This hardcoded `Gravatar` should be replaced with the Gravatar UserID when we implement the OAuth flow.

analytics/src/main/java/com/gravatar/analytics/tracks/TracksTracker.kt
AdamGrzybkowski: Do you know when we will have it?

hamorillo: Not yet. If there are no changes in the PR today, I'll ask for the review and an estimated time for the new release.

analytics/src/main/kotlin/com/gravatar/analytics/Event.kt
```diff
@@ -0,0 +1,5 @@
+package com.gravatar.analytics
+
+interface Event {
+    val name: String
```
AdamGrzybkowski, line 4: We will likely need a way to pass some params as well.

analytics/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
```
AdamGrzybkowski: I think we can remove the manifest if we don't use it?

analytics/src/test/kotlin/com/gravatar/analytics/di/AnalyticsModuleTest.kt
```diff
@@ -0,0 +1,23 @@
...
+
+class AnalyticsModuleTest : KoinTest {
+
+    @OptIn(KoinExperimentalAPI::class)
+    @Test
+    fun checkAllModules() {
+        analyticsModule.verify(
+            injections = injectedParameters(
+                definition<TracksTracker>(Context::class)
+            )
```
AdamGrzybkowski, line 20: When I comment out these two lines from the module
```kotiln
single { TracksClient.getClient(get()) }
singleOf(::TracksTracker) { bind<Tracker>() }
```
The test still passes, is that expected?

analytics/src/test/kotlin/com/gravatar/analytics/tracks/TrackTrackerTest.kt
```diff
@@ -0,0 +1,34 @@
...
+import io.mockk.mockk
+import io.mockk.verify
+import org.junit.Test
+
+class TrackTrackerTest {
+
+    @Test
+    fun `when trackEvent is invoked then call track on client`() {
+        val mockClient = mockk<TracksClient>(relaxed = true)
+        val tracker = TracksTracker(mockClient)
```
AdamGrzybkowski: 💡 Can we extract it and create in the `@Before` block?

app/src/main/java/com/gravatar/app/GravatarApplication.kt
```diff
@@ -12,7 +13,7 @@ class GravatarApplication : Application() {
 
         startKoin {
             androidContext(this@GravatarApplication)
-            modules(appModule)
+            modules(appModule, analyticsModule)
```
AdamGrzybkowski: ❓ I'm curious, is there a difference between this and adding the `analyticsModule` in the `appModule` via `includes()`?

AdamGrzybkowski (commented):
Looks good 👍 Left some clarifying comments.

analytics/src/test/kotlin/com/gravatar/analytics/di/AnalyticsModuleTest.kt
hamorillo: As far as I understand, if you comment out both lines, the module doesn't inject anything, so when the test verifies if any required dependency can't be resolved, nothing is missing (because nothing is being injected).

You can verify that the test is working by commenting just the first line:

```kotlin
single { TracksClient.getClient(get()) }
```

`TracksTracker` constructor requires a `TracksClient` instance, so if we don't explicitly specify how to instanciate that class, there will be an error during runtime. The test will catch that error.

```
* ----- > Missing definition for '[field:'tracksClient' - type:'com.automattic.android.tracks.TracksClient']' in definition '[Singleton: 'com.gravatar.analytics.tracks.TracksTracker',binds:com.gravatar.analytics.Tracker]'.
- Fix your Koin configuration -
1- add missing definition for type 'com.automattic.android.tracks.TracksClient'. like: singleOf(::com.automattic.android.tracks.TracksClient)
or
2- else define injection for 'com.gravatar.analytics.tracks.TracksTracker' with:
module.verify(
    injections = injectedParameters(
        definition<com.gravatar.analytics.tracks.TracksTracker>(com.automattic.android.tracks.TracksClient::class)
    )
)
```

app/src/main/java/com/gravatar/app/GravatarApplication.kt
hamorillo: That's a good question. Both ways should work. However, I see it as a question of definition. Also, I think it can have some impact on tests, as if `appModule` includes the `analyticsModule`, then `analyticsModule` will always be loaded.

However, I still have some doubts about the best approach for us after [reading the Koin docs](https://insert-koin.io/docs/reference/koin-core/modules/#module-includes-since-32).

analytics/src/test/kotlin/com/gravatar/analytics/tracks/TrackTrackerTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        tracker.trackEvent(event)
+
+        verify { mockClient.track(event.name, "test_user_id", TracksClient.NosaraUserType.ANON) }
+    }
+
+    @Test
+    fun `when flush is invoked then call flush on client`() {
+        tracker.flush()
+
+        verify(exactly = 1) { mockClient.flush() }
```
AdamGrzybkowski: We could replace this with 
```suggestion
         verifySequence {
            mockClient.flush()
        }
```

This way, if the `TrackeClient.flush()` is modified in a way that it tracks events after flushing, the test will fail.

analytics/src/test/kotlin/com/gravatar/analytics/di/AnalyticsModuleTest.kt
AdamGrzybkowski: Thanks for the explanation!

analytics/src/test/kotlin/com/gravatar/analytics/tracks/TrackTrackerTest.kt
hamorillo: 🤔 Do you want to check if we track an event inside the `flush` method by mistake?

app/src/main/java/com/gravatar/app/GravatarApplication.kt
AdamGrzybkowski: hmm looks like the end outcome is the same. I would lean towards including it in the `appModule` given that the `:app` glues the whole app, meaning all its dependencies.

analytics/src/test/kotlin/com/gravatar/analytics/tracks/TrackTrackerTest.kt
AdamGrzybkowski: Yeah, exactly. Just a suggestion, though.

hamorillo: I don't think of this as a failure point, but I applied your suggestion anyway. (It's not going to hurt :) )

app/src/main/java/com/gravatar/app/MainActivityViewModel.kt
hamorillo: **Update:** We'll need to use the WPCOM userName here.

AdamGrzybkowski (approved):
LGTM 👍  Thanks for all the updates!

PR #13, opened 2025-06-18
Launch OAuth flow to log in

AdamGrzybkowski:
### Description

This PR implements the first part of the OAuth flow. We are not requesting an actual token yet, only receiving the code from the redirect URI.

I've tried doing this directly via `deeplink` on the NavHost destination, but it didn't work as I would like to. Whenever the redirectUri would be opened in the app, the NavHost would completely recreate its backstack, meaning that pressing a back button would go back to the `splash` screen (the default destination). I couldn't find a way to change that behavior, so I've decided to use the same approach that we have in the SDK with a transparent activity to handle the deep link.

https://github.com/user-attachments/assets/29876b0d-8bdc-4699-961c-b392afbaff93

### Testing Steps

Add these two to the `local.properties` file:

```
oauth.clientId=
oauth.redirectUri=
```

1. Launch the app
2. Tap Log in
3. Try to dismiss, deny and approve the login
4. Confirm that the approval navigates to the home screen


loginUi/screenshotTests/roborazzi/com.gravatar.app.loginUi.presentation.login.LoginScreenTest.loginScreen_captureScreenshot.png
hamorillo, line 1: What's changed here? 🤔

app/build.gradle.kts
```diff
@@ -22,6 +22,7 @@ dependencies {
     implementation(project.dependencies.platform(libs.koin.bom))
     implementation(libs.koin.core)
     implementation(libs.koin.android)
+    implementation(libs.koin.compose)
```
hamorillo: ❓ I would say we are not using anything from that library. We may, but not for the moment, right?

hamorillo (approved):
Great work! 🚀

loginUi/screenshotTests/roborazzi/com.gravatar.app.loginUi.presentation.login.LoginScreenTest.loginScreen_captureScreenshot.png
AdamGrzybkowski: Hmm could it be that we have those created in different machines? I remember we had similar issues in the SDK.

app/build.gradle.kts
AdamGrzybkowski: You are right, I think! Must be a left over other solutions I tried.

loginUi/screenshotTests/roborazzi/com.gravatar.app.loginUi.presentation.login.LoginScreenTest.loginScreen_captureScreenshot.png
hamorillo: I don't think it's because those have been generated on different machines. I've just tested this branch and it works well in mine. 

Using the previous image I saw there are some changes:

image

I don't know why, though.

hamorillo: I see why:

image

Mystery solved.

PR #14, opened 2025-06-20
Add Roborazzi and integrate it into "loginUi" and "homeUi" modules

hamorillo:
### Description

Integrating Roborazzi to be able to implement screenshoot test to verify the UI components. 

This PR also creates a new module (`testUtils`), which will be used to share code for testing purposes between modules. In the current case, the Roborazzi configuration.

### Testing Steps

Modify locally either `HomeScreen` or `LoginScreen` composables and execute:

`./gradlew verifyRoborazziDebug -Pscreenshot`

- Verify that you get an error ❌ 

Revert the changes and execute the command again:

- Verify everything works ✅


testUtils/src/main/kotlin/com/gravatar/app/testUtils/roborazzi/RoborazziTest.kt
```diff
@@ -0,0 +1,46 @@
...
+            composable()
+        }
+        composeRule.onRoot().captureRoboImage()
+    }
+}
+
+/**
+ * You can filter ScreenshotTests using -Pscreenshot parameter
+ */
+interface ScreenshotTests
```
hamorillo, line 46: Exactly the same as we have in the [Gravatar-SDK-Repo](https://github.com/Automattic/Gravatar-SDK-Android). This allows us to filter only by the screenshot tests, not all unit tests.

build-logic/convention/src/main/kotlin/GravatarComposeConventionPlugin.kt
```diff
@@ -19,11 +20,39 @@ class GravatarComposeConventionPlugin : Plugin<Project> {
...
+                            if (project.hasProperty("screenshot")) {
+                                includeCategories(screenshotTestClass.name)
+                            } else {
+                                excludeCategories(screenshotTestClass.name)
+                            }
+                        } catch (_: ClassNotFoundException) {
+                            // ScreenshotTests class not found, skip configuration
+                            project.logger.info(
+                                "ScreenshotTests class not found, skipping screenshot test configuration"
+                            )
```
AdamGrzybkowski: ❓  I'm a bit confused about what this does. I did update the config to print started tests and when I run:
- `./gradlew testDebug` I could see both screenshots tests being started
- `./gradlew verifyRoborazziDebug -Pscreenshot` I can see other tests being started. `AppModuleTest` and another fake unit test that I've added just to test it.

Am I missing something about how this config should work?

hamorillo: Great catch @AdamGrzybkowski!

It wasn't working as I expected. More concretely:

```kotlin
val screenshotTestClass = Class.forName(
       "com.gravatar.app.testUtils.roborazzi.ScreenshotTests"
) 
``` 

My original idea was not to require all `Compose` modules to require the `testUtils`; however, I haven't found another straightforward approach. Also, I would say that we'll need screenshots in all modules defining composable anyway.

I've modified the approach and added the `testUtils` to the `app` module. This should work as expected.

AdamGrzybkowski (approved):
LGTM 👍 Thanks for fixing the filtering!

PR #17, opened 2025-06-24
Implement NavigationBar in the HomeScreen

hamorillo:
### Description

Implementing the NavigationBar in the home section with the three possible sections: Gravatar, Profile, and Share.

The main goal of the PR is to create the navigation and structure.

Alt Text

I've been checking some apps using the NavigationBar, and it looks like a common pattern to have a "Main" destination where the app will return to when the back button is pressed. If the back is pressed again, the app will be dismissed. The app should work in that way (this is, of course, open to discussion). Some examples are: Google Maps, WhatsApp, Gmail. 
It's also true that other apps, for example, Google Play, keep the whole back stack. 

### Testing Steps

- Run the app and verify that you switch between screens as expected.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
```diff
@@ -4,36 +4,124 @@ import androidx.compose.foundation.layout.Box
...
+    var selectedItem by rememberSaveable { mutableIntStateOf(0) }
+
+    DisposableEffect(navController) {
+        val listener = NavController.OnDestinationChangedListener { _, destination, _ ->
+            selectedItem = when (destination.route) {
+                GravatarDest::class.qualifiedName -> 0
+                ProfileDest::class.qualifiedName -> 1
+                ShareDest::class.qualifiedName -> 2
+                else -> selectedItem
+            }
```
hamorillo: I'm not super happy with this piece of code; any suggestions will be welcomed.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/navigation/HomeNavigation.kt
```diff
@@ -0,0 +1,39 @@
...
+import kotlinx.serialization.Serializable
+
+@Serializable
+data object GravatarDest
+
+@Serializable
+data object ProfileDest
+
+@Serializable
+data object ShareDest
```
AdamGrzybkowski: Could those be `internal`?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/navigation/HomeNavigation.kt
```diff
@@ -0,0 +1,39 @@
...
+
+@Composable
+fun HomeNavigation(
```
AdamGrzybkowski: This as well, I think?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
AdamGrzybkowski: Could we set the `selectedItem` to a current one in each `onClick`? That would remove this whole `DisposableEffect`. 

```kotlin
                     onClick = {
                        if (selectedItem != 0) {
                            navController.navigate(GravatarDest) {
                                popUpTo(navController.graph.startDestinationId)
                                launchSingleTop = true
                            }
                            selectedItem = 0
                        }
                    }
```

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
```diff
@@ -1,41 +1,118 @@
...
+                            painterResource(id = R.drawable.gravatar),
+                            contentDescription = stringResource(R.string.gravatar)
+                        )
+                    },
+                    label = { Text(stringResource(R.string.gravatar)) },
+                    selected = selectedItem == 0,
+                    onClick = {
+                        if (selectedItem != 0) {
+                            navController.navigate(GravatarDest) {
+                                popUpTo(navController.graph.startDestinationId)
```
AdamGrzybkowski: This is so that we don't switch between the screens when using the system back button, correct? 

❓ Do you know how this will behave when we start having navigation inside each of those bottom tab destinations?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
```diff
@@ -1,41 +1,118 @@
...
+                    selected = selectedItem == 1,
+                    onClick = {
+                        if (selectedItem != 1) {
+                            navController.navigate(ProfileDest) {
+                                popUpTo(navController.graph.startDestinationId)
+                                launchSingleTop = true
+                            }
+                        }
+                    }
+                )
```
AdamGrzybkowski: 💡 We could have a sealed class for all these home destinations and iterate over them here rather them adding each `NavigationBarItem` manually.

homeUi/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="gravatar">Gravatar</string>
+    <string name="share">Share</string>
+    <string name="profile">Profile</string>
```
AdamGrzybkowski: Given that those are specific labels for each tab item, maybe we should prefix those strings so that they won't be used accidentally?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
AdamGrzybkowski: Something like this:

```kotlin
@Serializable
sealed class HomeDestination(
    val iconRes: Int,
    val labelRes: Int,
    val position: Int,
) {
    @Serializable
    data object Gravatar : HomeDestination(
        iconRes = com.gravatar.app.homeUi.R.drawable.gravatar,
        labelRes = com.gravatar.app.homeUi.R.string.gravatar,
        position = 0
    )

    @Serializable
    data object Profile : HomeDestination(
        iconRes = com.gravatar.app.homeUi.R.drawable.profile_icon,
        labelRes = com.gravatar.app.homeUi.R.string.profile,
        position = 1
    )

    @Serializable
    data object Share : HomeDestination(
        iconRes = com.gravatar.app.homeUi.R.drawable.qr_code_icon,
        labelRes = com.gravatar.app.homeUi.R.string.share,
        position = 2
    )

    companion object {
        val allDestinations = listOf(Gravatar, Profile, Share)
    }
}
```

and then inside the `NavigationBar`

```kotlin
                  HomeDestination.allDestinations
                    .sortedBy { it.position }
                    .forEach { destination ->
                        NavigationBarItem(
                            icon = {
                                Icon(
                                    painterResource(id = destination.iconRes),
                                    contentDescription = stringResource(destination.labelRes)
                                )
                            },
                            label = { Text(stringResource(destination.labelRes)) },
                            selected = selectedItem == destination.position,
                            onClick = {
                                if (selectedItem != destination.position) {
                                    navController.navigate(destination) {
                                        popUpTo(navController.graph.startDestinationId)
                                        launchSingleTop = true
                                    }
                                }
                            }
                        )
                    }
```

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
AdamGrzybkowski: Not sure how the back button behavior would change in this case.

AdamGrzybkowski (commented):
Works great, but left some suggestions and ideas for the open question.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
hamorillo: That was my first approach, and what you can see in the majority of examples. However, that doesn't work well with the back button to keep the selection in the NavigationBar :

https://github.com/user-attachments/assets/ab1d9414-c20f-410b-8052-9246b8522db1

The `selectedItem`, which is used to highlight the selected item, won't be changed. Therefore, we still need a way to detect when the current item has been changed (due to the navigation handling the back button) and update it accordingly.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
hamorillo: > This is so that we don't switch between the screens when using the system back button, correct?

Yes.

> Do you know how this will behave when we start having navigation inside each of those bottom tab destinations?

I'm not entirely sure, but I would say we'll need to utilize the `saveState` and `restoreState` flags to preserve the previous state.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
hamorillo: Nice suggestion. Thanks! 👏

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
AdamGrzybkowski: Right, I think I found another solution:

```kotlin
    val backStackEntry = navController.currentBackStackEntryAsState() <-- new line
    Scaffold(
        ...
           .forEach { destination ->
               NavigationBarItem(
                  ...
                  selected = destination.route == backStackEntry.value?.destination?.route, <-- modified
                  onClick = {
                       if (backStackEntry.value?.destination?.route != destination.route) { <-- modified
                            navController.navigate(destination) {
                            popUpTo(navController.graph.startDestinationId)
                            launchSingleTop = true
                       }
                  ...          
```

I did a quick test, and it seems to work just fine without the listener.

AdamGrzybkowski (approved):
Works great! I've left one suggestion regarding the current destination listener.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
hamorillo: It works well, and I haven't found a reason not to use this approach. Since it's simpler than mine, I've made the changes. :) 

Thanks for your work. 🙏

PR #18, opened 2025-06-25
Implement `ProfileHeader` Composable

hamorillo:
### Description

The main goal of this PR is to implement the ProfileHeader Composable. We need to add the [Gravatar SDK](https://github.com/Automattic/Gravatar-SDK-Android) dependencies for two reasons:

- Fetch the profile of a given user
- Use some of the UI components provided by the library (in this case, the `Avatar` Composable)

image

**Note:** I've hardcoded my user for the moment. Once the OAuth tasks are complete, we'll need to use that information to fetch the profile.

### Testing Steps

- Run the app and verify you can see the `ProfileHeader` in the Profile section.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileHeader.kt
```diff
@@ -0,0 +1,123 @@
...
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.gravatar.extensions.defaultProfile
+import com.gravatar.restapi.models.Profile
+import com.gravatar.ui.components.atomic.Avatar
+
+@Composable
+fun ProfileHeader(
+    profile: Profile,
```
hamorillo, line 28: I'm using the `Profile` class from the Gravatar-SDK. If that were from a third-party library, it would be a good idea to map that object to another one managed by the app. In this case, since we are using our library, I consider it under our control.

homeUi/screenshotTests/roborazzi/com.gravatar.app.homeUi.presentation.home.profile.ProfileHeaderTest.profileHeader_bothJobTitleAndCompany.png
AdamGrzybkowski, line 1: 💡 We can set a fake image loader for the screenshot test like we did in the SDK - https://github.com/Automattic/Gravatar-SDK-Android/blob/trunk/uitestutils/src/main/java/com/gravatar/uitestutils/RoborazziTest.kt#L60

Not a big deal, but it helps visualize the full component better.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileHeader.kt
```diff
@@ -0,0 +1,110 @@
...
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.gravatar.extensions.defaultProfile
+import com.gravatar.restapi.models.Profile
+import com.gravatar.ui.components.atomic.Avatar
+
+@Composable
+fun ProfileHeader(
```
AdamGrzybkowski: ⛏️ Missing `internal` modifier

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
```diff
@@ -2,17 +2,34 @@ package com.gravatar.app.homeUi.presentation.home.profile
...
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import org.koin.androidx.compose.koinViewModel
 
 @Composable
-fun ProfileScreen() {
+fun ProfileScreen(viewModel: ProfileViewModel = koinViewModel()) {
```
AdamGrzybkowski: We can make this `internal` as well.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
```diff
@@ -2,17 +2,34 @@ package com.gravatar.app.homeUi.presentation.home.profile
...
+        } else {
+            uiState.profile.let { profile ->
+                if (profile != null) {
+                    ProfileHeader(profile = profile, modifier = Modifier.padding(16.dp))
+                } else {
+                    Text("There was an error retrieving the profile.")
+                }
+            }
+        }
     }
```
AdamGrzybkowski, line 41: 💡 I think we should extract this to a new `ProfileScreen(uiState: ProfileUiState)` component - otherwise it's hard to create previews or run screenshot tests for the whole screen.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileHeader.kt
```diff
@@ -0,0 +1,110 @@
...
+            )
+
+            profile.jobInfo().takeIf { it.isNotBlank() }?.let { jobInfo ->
+                Text(
+                    text = jobInfo,
+                    style = MaterialTheme.typography.bodyMedium,
+                    color = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.6f)
+                )
+            }
+        }
```
AdamGrzybkowski: 💡 We have a similar problem here like in the SDK - the card doesn't look good when you start scaling the font size. What if we used the `autoSize` and set those to single lines? Something like:

```kotlin
            BasicText(
                text = profile.displayName,
                style = MaterialTheme.typography.titleMedium,
                maxLines = 1,
                autoSize = TextAutoSize.StepBased(
                       maxFontSize = 16.sp
                 )
            )
```

It needs some styling but here's the difference with the max font scaling

| Before | After |
|---|---|
| Screenshot 2025-06-25 at 14 44 53 | Screenshot 2025-06-25 at 14 44 45 |

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
```diff
@@ -2,17 +2,34 @@ package com.gravatar.app.homeUi.presentation.home.profile
...
         modifier = Modifier.fillMaxSize(),
-        contentAlignment = Alignment.Center
     ) {
-        Text("Profile Screen")
+        if (uiState.isLoading) {
+            CircularProgressIndicator(modifier = Modifier.align(Alignment.Center))
+        } else {
+            uiState.profile.let { profile ->
+                if (profile != null) {
+                    ProfileHeader(profile = profile, modifier = Modifier.padding(16.dp))
```
AdamGrzybkowski, line 35: ❓ This will have to be a top bar. I wonder if we will need a scaffold here as well, or we will just keep it at the top? Have you thought about this?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileHeader.kt
AdamGrzybkowski: Yeah, I have mixed feelings about this 😬, but the components expect the same models so it is how it is.

AdamGrzybkowski (approved):
Looks good 👍 I've left some comments, but none are breaking.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
hamorillo: I haven't thought much about it, as we'll implement the animation to collapse/expand the header, and that can require some changes. My idea was to keep it at the top for now.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
hamorillo: Done! Also added a screenshot test.

homeUi/screenshotTests/roborazzi/com.gravatar.app.homeUi.presentation.home.profile.ProfileHeaderTest.profileHeader_bothJobTitleAndCompany.png
hamorillo: Yeah, I was considering this, but if you don't mind, I'll open a different PR with it.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileHeader.kt
hamorillo: I've applied this suggestion. 🙂

PR #19, opened 2025-06-25
OAuth - request the token via REST API

AdamGrzybkowski:
### Description

This PR adds the logic for requesting the user's Token via WP REST API after the app receives the `code` from the OAuth flow. 

Some new stuff that's introduced:
- The `:foundations` module. This is the place for the lowest-level types that we will share between the modules. One example could be a custom Result type that should be common for all the modules. 
- The `:userComponent` module. This is the place for user-related domain/data stuff. At this point, we only do the auth, but I believe this is also the place where we can put stuff like getting the user profile. 

Let me know what you think about it 🙏 

There are no tests, the error handling is not done, but I don't want this PR to grow, so I would rather do this stuff separately, once we have the full OAuth flow completed. 

### Testing Steps

1. Launch the app and login. 
2. You can verify that the app won't crash with a network error (wrong OAuth setup, for example), but in general, the error handling is not there except for not crashing and showing a Toast.
3. Verify the success flow - you can monitor the network and see if it works.


loginUi/build.gradle.kts
```diff
@@ -32,6 +32,11 @@ android {
                 "OAUTH_REDIRECT_URI",
                 "\"${properties["oauth.redirectUri"]?.toString() ?: ""}\"",
             )
+            buildConfigField(
+                "String",
+                "OAUTH_CLIENT_SECRET",
+                "\"${properties["oauth.clientSecret"]?.toString() ?: ""}\"",
+            )
```
AdamGrzybkowski, line 39: Remember to add this to your `local.properties`

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/HttpClientModule.kt
```diff
@@ -0,0 +1,22 @@
...
+        HttpClient(OkHttp) {
+            install(ContentNegotiation) {
+                json(
+                    Json {
+                        ignoreUnknownKeys = true
+                    }
+                )
+            }
+        }
+    }
```
AdamGrzybkowski: We can configure more stuff here, so we don't have to repeat it in each request, but let's do that as we need it. This is also the place to configure stuff like network loggers if we need to in the future.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/UserComponentModule.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.gravatar.app.usercomponent.data.RealAuthRepository
+import com.gravatar.app.usercomponent.data.WordPressClient
+import com.gravatar.app.usercomponent.domain.repository.AuthRepository
+import org.koin.core.module.dsl.bind
+import org.koin.core.module.dsl.factoryOf
+import org.koin.dsl.module
+
+val userComponentModule = module {
+    factoryOf(::RealAuthRepository) { bind<AuthRepository>() }
+    factoryOf(::WordPressClient)
```
AdamGrzybkowski, line 12: I don't see a reason that those should be `singletons`. WDYT?

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/WordPressClient.kt
```diff
@@ -0,0 +1,63 @@
...
+        private const val BASE_URL = "https://public-api.wordpress.com"
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    suspend fun login(
+        code: String,
+        clientSecret: String,
+        redirectUri: String,
+        clientId: String
+    ): Result<String> = withContext(dispatcherProvider.io) {
```
AdamGrzybkowski, line 30: Standard `Result` for now - but I want to introduce our own type for better error types, rather than exceptions.

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginScreen.kt
```diff
@@ -7,38 +7,79 @@ import androidx.compose.foundation.layout.Column
...
     onLoggedIn: () -> Unit,
 ) {
+    LoginScreen(
+        onLoggedIn = onLoggedIn,
+        viewModel = koinViewModel(),
+    )
+}
+
+@Composable
+internal fun LoginScreen(
```
hamorillo, line 41: Preview in this class no longer works, as Koin won't be initialized. [We could use KoinAppApplicationPreview:](https://insert-koin.io/docs/reference/koin-compose/compose/#compose-preview-with-koinapplicationpreview)

```kotlin
@Preview
@Composable
private fun LoginScreenPreview() {
    KoinApplicationPreview(application = { modules(loginUiModule) }) {
        LoginScreen(
            onLoggedIn = { },
        )
    }
}
```

However, if you try that code, it will fail because it doesn't know how to resolve `dispatcherModule`. 

`dispatcherModule` is being defined in the `app` module, so `DispatcherProvider` will be available everywhere in the app when needed. However, thinking in isolated modules, `app`module doesn't require `:foundations`. It's only required by `loginUI`. 

I wonder if `dispatcherModule` should be defined in `foundations` and refactor a bit the dependencies.

Something like this:
https://github.com/Automattic/Gravatar-Android/compare/adam/GRA-199...hamorillo/GRA-199-refactor-dispatcher

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/UserComponentModule.kt
hamorillo: Agree. We are preserving data in those instances, so no need for a singleton. 👌

hamorillo (commented):
Works as described. I can see the `access_token` when reviewing the network requests. 

I left a comment about DI and modules that could be worth discussing before approving it.

> There are no tests, the error handling is not done, but I don't want this PR to grow, so I would rather do this stuff separately, once we have the full OAuth flow completed.

Sounds good. I would also include in those tests the `Koin.Modules` ones. :)

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginScreen.kt
AdamGrzybkowski: This should be fixed now. I didn't update the test to use the proper composable that doesn't rely on ViewModel, but rather on a ui state class.

hamorillo (approved):

PR #21, opened 2025-06-26
Auth token storage

AdamGrzybkowski:
### Description

This PR introduced the auth token storage backed by Datastore preferences. It's part of the `:userComponent` modules and is completely internal at this point. If we see that the token is required by more component, we will likely have to extract this.

To keep things simple, I've added the `AuthRepository` directly in some composables to log out the user and check the user session.

### Testing Steps
1. Launch the app
2. Log in
3. Confirm you are on the home screen
4. Kill the app
5. Launch the app
6. Confirm you are on the home screen
7. Log out 
8. Confirm you are on the login screen


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealAuthRepository.kt
```diff
@@ -13,8 +14,17 @@ internal class RealAuthRepository(
...
-            onSuccess = { Result.success(Unit) },
+            onSuccess = { token ->
+                tokenStorage.save(token)
+                Result.success(Unit)
+            },
             onFailure = { Result.failure(it) }
         )
     }
+
+    override suspend fun isUserLoggedIn(): Boolean = tokenStorage.get() != null
```
hamorillo: 💡 Since we are already using `DataStorage`, which provides `Flows`, we may also consider exposing a Flow here instead of implementing a one-time call function.

With that approach, for example, at some point the token expires; we can handle the navigation to the login screen in just one place, without needing to consider those cases everywhere.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealAuthRepository.kt
```diff
@@ -13,8 +14,17 @@ internal class RealAuthRepository(
             redirectUri = loginRequest.redirectUri,
             clientId = loginRequest.clientId
         ).fold(
-            onSuccess = { Result.success(Unit) },
+            onSuccess = { token ->
+                tokenStorage.save(token)
```
hamorillo, line 20: We are saving the token in plain text without any encryption. However, I think this should be enough (at least for now) as it will be saved locally on the user's device.

We can evaluate some encryption later if needed.

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/di/UserComponentModuleTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.koin.test.verify.verify
+
+class UserComponentModuleTest : KoinTest {
+
+    @OptIn(KoinExperimentalAPI::class)
+    @Test
+    fun checkAllModules() {
+        userComponentModule.verify(
+            injections = injectedParameters(
+                definition<DatastoreAuthTokenStorage>(DispatcherProvider::class),
```
hamorillo:
```suggestion

```

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/di/DatastoreModuleTest.kt
```diff
@@ -0,0 +1,27 @@
...
+class DatastoreModuleTest : KoinTest {
+
+    @OptIn(KoinExperimentalAPI::class)
+    @Test
+    fun checkDatastoreModule() {
+        datastoreModule.verify(
+            injections = injectedParameters(
+                definition<DataStore<Preferences>>(Context::class),
+                definition<DatastoreAuthTokenStorage>(DispatcherProvider::class)
+            )
```
hamorillo:
```suggestion
        datastoreModule.verify()
```

hamorillo (commented):
Works as described! I left some comments / ideas.

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/di/DatastoreModuleTest.kt
AdamGrzybkowski: Done

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/di/UserComponentModuleTest.kt
AdamGrzybkowski: Done

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealAuthRepository.kt
AdamGrzybkowski: Good idea! I pushed the changes - the `isUserLoggedIn` is not collected in the Root, and we don't need the callback for loggedIn/loggedOut anymore.

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreenTest.kt
```diff
@@ -8,9 +10,11 @@ class HomeScreenTest : RoborazziTest() {
...
     fun homeScreen() {
         screenshotTest {
-            HomeScreen(
-                onLoggedOut = { }
-            )
+            HomeScreen {
+                ProfileScreen(
+                    uiState = ProfileUiState()
+                )
+            }
```
hamorillo, line 17: 👍 I suppose we'll need to refine our test when those screens are finalized.

hamorillo (approved):
Thanks for applying the `flow` refactor! Looks great to remove all those callbacks from the code :) 🚀

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreenTest.kt
AdamGrzybkowski: Yeah, but I think the main point in this one is testing the bottom bar look, so that's covered.

PR #22, opened 2025-06-26
Add Profile Fields in Profile Tab - Only visualization, not edition

hamorillo:
### Description

This PR adds the Profile Fields composables and utilizes them in the Profile Tab. We are reusing them from the SDK as the functionality is very similar to the QE.

This PR only handles the visualization of the data, not the editing. In a following PR, we'll implement everything needed to allow the user to modify any data from that screen.

https://github.com/user-attachments/assets/72772775-bc3b-459d-a7a6-fa9df5aaee3d

**Note:** The contact section is missing. I've implemented it in a [separate branch](https://github.com/Automattic/Gravatar-Android/compare/hamorillo/GRA-217...hamorillo/GRA-217-contact-section), as it won't be populated until we make an authenticated request.

### Testing Steps

- Run the app and verify that the profile fields look as expected.
- Verify the state of the Profile Tabs is saved when switching between tabs.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
```diff
@@ -42,8 +42,11 @@ fun HomeScreen(
                             onClick = {
                                 if (backStackEntry.value?.destination?.route != destination.route) {
                                     navController.navigate(destination) {
-                                        popUpTo(navController.graph.startDestinationId)
+                                        popUpTo(navController.graph.startDestinationId) {
+                                            saveState = true
+                                        }
                                         launchSingleTop = true
+                                        restoreState = true
```
hamorillo, line 57: We need to save the state when switching between tabs.

homeUi/screenshotTests/roborazzi/com.gravatar.app.homeUi.presentation.home.profile.header.ProfileHeaderTest.profileHeader_bothJobTitleAndCompany.png
hamorillo, line 1: 🔔  This screenshot (and the rest from the same file) are identical but have been moved to a different directory. That's why they are marked here.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/about/AboutFieldsSection.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+@Composable
+internal fun AboutFieldsSection(
+    label: String?,
+    fields: Set<AboutEditorField>,
+    formEnabled: Boolean,
+    onValueChange: (AboutEditorField) -> Unit,
```
hamorillo, line 19: I kept some code that is not being used in this PR, but will likely be used in the next one when enabling the edition.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
```diff
@@ -32,7 +39,23 @@ internal fun ProfileScreen(uiState: ProfileUiState) {
         } else {
             uiState.profile.let { profile ->
                 if (profile != null) {
-                    ProfileHeader(profile = profile, modifier = Modifier.padding(16.dp))
+                    Column {
+                        Spacer(modifier = Modifier.height(16.dp))
+
+                        ProfileHeader(profile = profile, modifier = Modifier.padding(horizontal = 16.dp))
+
+                        Spacer(modifier = Modifier.height(16.dp))
```
AdamGrzybkowski:
```suggestion
                        ProfileHeader(profile = profile, modifier = Modifier.padding(16.dp))
```

⛏️ I believe this would result in the same UI, any reason to go with spacers here?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileViewModel.kt
```diff
@@ -2,22 +2,29 @@ package com.gravatar.app.homeUi.presentation.home.profile
...
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 
 class ProfileViewModel(private val profileService: ProfileService) : ViewModel() {
```
AdamGrzybkowski: ⛏️ Not part of this PR but we are missing the `internal` modifier here.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileViewModel.kt
```diff
@@ -38,4 +48,42 @@ class ProfileViewModel(private val profileService: ProfileService) : ViewModel()
...
+
+    private fun profileObserver() {
+        uiState
+            .distinctUntilChanged { old, new -> old.profile == new.profile }
+            .onEach {
+                _uiState.update { currentState ->
+                    currentState.copy(aboutFields = currentState.profile?.aboutFields() ?: emptySet())
+                }
+            }
+            .launchIn(viewModelScope)
```
AdamGrzybkowski: ❓ Curious, why like this and not simply when we get the profile above? Not a big deal, but we're updating the state twice instead of once.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/about/AboutInputField.kt
```diff
@@ -0,0 +1,122 @@
+package com.gravatar.app.homeUi.presentation.home.profile.about
+
+/**
+ * Represents a set of fields that can be shown in the "About" section.
+ *
+ * @property value The value of the about input field.
+ */
+internal class AboutInputField(
+    val value: String,
+) {
```
AdamGrzybkowski: 💡 We can use a sealed class here (or enums if that's enough). We couldn't in the SDK to not lock ourselves to a fixed nr of AboutInputFields.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/about/AboutSection.kt
```diff
@@ -0,0 +1,154 @@
...
+private val Set<AboutEditorField>.professionalLabelRes: Int?
+    @StringRes get() = if (hasOtherSectionsThan(Section.PROFESSIONAL)) {
+        R.string.about_field_section_label_professional
+    } else {
+        null
+    }
+
+private fun Set<AboutEditorField>.hasOtherSectionsThan(other: Section): Boolean {
+    return any { it.type.section != other }
+}
```
AdamGrzybkowski: 💡 We display all the sections in the app, so we don't need this logic. In the SDK, the number of fields shown was dynamic, so some sections could be completely hidden, hence this. The section label is always visible from what I understand.

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileViewModelTest.kt
```diff
@@ -0,0 +1,109 @@
...
+                    ProfileUiState(isLoading = false, profile = mockProfile, aboutFields = mockProfile.aboutFields()),
+                    awaitItem()
+                )
+            }
+        }
+
+    @Test
+    fun `when profile is available then aboutFields are correctly populated`() {
+        // Given
+        val mockProfile = mockProfile()
```
AdamGrzybkowski:
```suggestion
        val profile = defaultProfile()
```

Could we rename? This is not a mock and it confused me a bit :D

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileViewModelTest.kt
```diff
@@ -0,0 +1,109 @@
...
+import org.junit.Test
+import java.net.URI
+
+class ProfileViewModelTest {
+    private val testDispatcher = StandardTestDispatcher()
+
+    @get:Rule
+    var coroutineTestRule = CoroutineTestRule(testDispatcher)
+
+    private val profileService = mockk<ProfileService>(relaxed = true)
```
AdamGrzybkowski:
```suggestion
    private val profileService = mockk<ProfileService>(relaxed = false)
```

I think it's better to always explicitly mock what we need, and to have a crash if we miss it. This can lead to 🟢  test by accident.

AdamGrzybkowski (commented):
Works good! I left some comments related to the code that we can make better here without the restrictions we have in the SDK.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileViewModel.kt
hamorillo: Now the Profile is only modified when fetching; however, in the next iteration, it will also be modified when the user edits any of the fields. This way, we are sure we don't forget to update the `aboutFields`.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/about/AboutInputField.kt
hamorillo: 👍 I've refactored this into an enum class.

hamorillo (commented):
I've applied your suggestions.

AdamGrzybkowski (approved):
:shipit: 


EDIT: Approved, but forgot to add a comment about the `profileObserver` - take a look before merging 🙏

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileViewModel.kt
AdamGrzybkowski: You could just do this:
```kotlin
internal data class ProfileUiState(
    val isLoading: Boolean = false,
    val profile: Profile? = null,
) {

    val aboutFields: Set<AboutEditorField> = profile.aboutFields()
}
```

I don't think we need to set those separately. The drawback here is that the `aboutFields` is not part of the `equals` check, but in this case that's okay, because verifying the `profile` is enough (assuming you will update the `profile` on input field edits, not the `aboutFields` directly).

PR #23, opened 2025-06-30
Added Contact section to profile screen

hamorillo:
### Description

Adding the contact section to the Profile screen. 

image

**Note:** Contact files are only returned when making an authenticated request.

### Testing Steps

- Launch the app
- Switch to the Profile tab
- Verify you can see the Contact fields


AdamGrzybkowski (approved):
:shipit:

PR #24, opened 2025-06-30
Show Avatar grid in the Gravatar tab

AdamGrzybkowski:
### Description

This PR adds the avatars grid in the Gravatar tab. To fetch the avatars, I created `UserRepository` that will handle all user-related operations. 

I wrapped the whole tab in `PullToRefreshBox` so that the grid can be refreshed. 

No proper error handling for now, just making sure there's no crash. 

Screenshot 2025-06-30 at 11 52 14

### Testing Steps

1. Launch the app
2. Log in
3. Confirm you see the avatars


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
```diff
@@ -0,0 +1,36 @@
...
+internal class RealUserRepository(
+    private val avatarService: AvatarService,
+    private val profileService: ProfileService,
+    private val tokenStorage: AuthTokenStorage,
+) : UserRepository {
+
+    override suspend fun getAvatars(): Result<List<Avatar>> {
+        val token = tokenStorage.get().first()
+        return if (token != null) {
+            val avatars = profileService.retrieveAuthenticatedCatching(token)
```
AdamGrzybkowski, line 19: We need to get the hash in order to fetch the avatars. This is here for now, but we will have to move this call and make sure we have all that info already after the user logs in.

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/data/RealUserRepositoryTest.kt
```diff
@@ -0,0 +1,128 @@
...
+            tokenStorage.save(testToken)
+
+            // Mock profile service
+            val profileResult = GravatarResult.Success<Profile, ErrorType>(profile)
+            coEvery { profileService.retrieveAuthenticatedCatching(testToken) } returns profileResult
+
+            // Mock avatar service
+            val avatarResult = GravatarResult.Success<List<Avatar>, ErrorType>(avatars)
+            coEvery {
+                avatarService.retrieveCatching(testToken, any<Hash>())
```
AdamGrzybkowski: `Hash` doesn't have `.equals` nor `.hashCode` functions so the matchers don't work well here and I had to use `any`.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/AvatarsGridSection.kt
```diff
@@ -0,0 +1,125 @@
...
+                modifier = Modifier,
+            )
+        }
+    }
+}
+
+@Composable
+@Preview(showBackground = true)
+private fun VerticalAvatarSectionPreview() {
+    GravatarTheme {
```
hamorillo: This `GravatarTheme` is provided by the SDK and not the App's one. As the one from the app won't be accessible here, we should probably consider extracting it to a different module. (In a different task / PR).

In this case, I would keep just `MaterialTheme`:
```suggestion
    MaterialTheme {
```

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarUiState.kt
```diff
@@ -0,0 +1,25 @@
...
+            isLoading = false,
+        )
+    }
+}
+
+internal sealed class AvatarUi(val avatarId: String) {
+    data class Uploaded(
+        val avatar: Avatar,
+        val isSelected: Boolean,
+        val isLoading: Boolean,
```
hamorillo, line 23: ❓ Will this be used when we implement the selection?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+                        currentState.copy(avatars = avatars)
+                    }
+                }
+                .onFailure {
+                }
+            _uiState.update { currentState ->
+                currentState.copy(
+                    isRefreshing = false,
+                    isLoading = false
+                )
```
hamorillo: 🤔 Not a big deal, but I would expect this update to be included in the `onSuccess` or `onFailure` uiStateUpdates.

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/data/RealUserRepositoryTest.kt
hamorillo: This should do the trick:

```suggestion
                avatarService.retrieveCatching(testToken, match { it.toString() == testHash })
```

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/data/RealUserRepositoryTest.kt
```diff
@@ -0,0 +1,128 @@
...
+            // When
+            val result = repository.getAvatars()
+
+            // Then
+            assertTrue(result.isSuccess)
+            assertEquals(avatars, result.getOrNull())
+
+            // Verify interactions
+            coVerify { profileService.retrieveAuthenticatedCatching(testToken) }
+            coVerify { avatarService.retrieveCatching(testToken, any<Hash>()) }
```
hamorillo, line 78: Same here:

```suggestion
            coVerify { avatarService.retrieveCatching(testToken, match { it.toString() == testHash }) }
```

hamorillo (commented):
Works well! I left some minor comments.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarUiState.kt
AdamGrzybkowski: Yes, this is used when the request to select the avatar is in progress.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModel.kt
AdamGrzybkowski: Done

hamorillo (approved):
🚀

PR #25, opened 2025-07-01
Select an avatar action

AdamGrzybkowski:
### Description

This PR adds an option to select an avatar. I used the same Popup style contextual menu as in the SDK for the avatar options. For now, only `Select` one works; the other is a temp one.

I've also added the ✅ icon on top of the selected avatar.

There are no Snackbars for confirmation/errors yet.

We will have to polish this a bit and lock selecting the currently selected avatar, or when another API request is in progress.

https://github.com/user-attachments/assets/277417cd-de84-4ad1-93cb-b1eddce4d019

### Testing Steps

1. Launch the app
2. Wait for the avatars to load
3. Tap on the avatar to open the menu
4. Tap Select
5. Confirm the avatar was selected
6. Try without the internet


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
```diff
@@ -14,6 +15,28 @@ internal class RealUserRepository(
     private val tokenStorage: AuthTokenStorage,
 ) : UserRepository {
 
+    override suspend fun selectAvatar(avatarId: String): Result<Unit> {
+        val token = tokenStorage.get().first()
```
hamorillo:
```suggestion
        val token = tokenStorage.get().firstOrNull()
```

hamorillo (approved):
Looks good! Left a minor comment.

PR #26, opened 2025-07-01
Update Profile Fields

hamorillo:
### Description

This PR enables profile field editing and allows users to save changes in their Gravatar account.

https://github.com/user-attachments/assets/cc85a4a4-c57b-420c-a80b-ca40187a6462

The "Save" button will require refactoring with the final designs. 

**Note:** Errors aren't shown to the user in the UI. We'll need design or at least add some toasts.

### Testing Steps

- Launch the app
- Navigate to the Profile tab
- Verify you can edit any field and save it
- Verify that if you close the app and open it again, the updated fields are loaded (so the change has landed in the backend).


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
```diff
@@ -36,7 +37,7 @@ internal class RealUserRepository(
     }
 
     override suspend fun getProfile(): Result<Profile> {
-        val token = tokenStorage.get().first()
+        val token = tokenStorage.get().firstOrNull()
```
hamorillo, line 63: Copilot detected this (I had the same in the new method), which I would say is a good catch.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileUiState.kt
```diff
@@ -1,12 +1,26 @@
...
+    val originalAboutFields: Set<AboutEditorField> = profile?.aboutFields() ?: emptySet()
+
+    val aboutFields: Set<AboutEditorField> = originalAboutFields.map { field ->
+        val editedValue = editedAboutFields[field.type]
+        if (editedValue != null) {
+            field.copy(value = editedValue)
+        } else {
+            field
+        }
+    }.toSet()
```
AdamGrzybkowski, line 23: Nice, I like this!

AdamGrzybkowski (commented):
Looks good, saving work as described.

One thing I noticed - you can't press the `save` button with the keyboard visible after you tap on one of the fields at the bottom of the screen. The keyboard pushes the header out completely or partially and you can bring it back without hiding the keyboard.
header_with_keyboard

hamorillo:
> One thing I noticed - you can't press the save button with the keyboard visible after you tap on one of the fields at the bottom of the screen. The keyboard pushes the header out completely or partially and you can bring it back without hiding the keyboard.

It should be fixed now. Thanks!

AdamGrzybkowski (approved):
:shipit:

PR #27, opened 2025-07-01
Upload avatar via camera

AdamGrzybkowski:
### Description

This PR adds two new buttons to upload an avatar either via Camera or from the photo gallery. 

The `Photo gallery` one doesn't work yet, this will be done in a separate PR. 

The code is mostly taken from the SDK, so the flow is the same.

https://github.com/user-attachments/assets/71489b17-0260-4631-80e5-fc7584b50af5

### Testing Steps

1. Launch the app
2. Tap `Camera` button
3. Take a picture and crop it
4. Confirm the image is being uploaded and added to the list


app/src/test/kotlin/com/gravatar/app/di/AppModuleTest.kt
```diff
@@ -10,6 +11,8 @@ class AppModuleTest : KoinTest {
     @OptIn(KoinExperimentalAPI::class)
     @Test
     fun checkAllModules() {
-        appModule.verify()
+        appModule.verify(
+            extraTypes = listOf(Context::class)
```
AdamGrzybkowski, line 15: This test is failing because it has a missing `Context` type for the `FileUtils` constructor. The problem is only here in the test, so I assume that should be fine, but if you know about a better way to handle it, let me know!

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/FileUtils.kt
```diff
@@ -0,0 +1,19 @@
+package com.gravatar.app.homeUi.presentation
+
+import android.content.Context
+import android.net.Uri
+import androidx.core.net.toFile
+import java.io.File
+
+internal class FileUtils(
```
hamorillo, line 8: 🤔 I'm not sure if this file is in the best possible location. It looks a bit odd to have it under `homeUi/presentation`. 

I would say this could look better in a different module, but on the other hand, I'm not sure if the work is worth it. 

Just thinking aloud, so feel free to skip the comment.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/UploadAvatarSection.kt
```diff
@@ -0,0 +1,50 @@
...
+            NewAvatarButton(
+                label = stringResource(R.string.gravatar_tab_photos),
+                iconRes = R.drawable.ic_photo_library,
+                contentDescription = stringResource(R.string.gravatar_tab_photos_content_description),
+                onClick = onChooseFromGalleryClicked,
+                modifier = Modifier.weight(1f)
+            )
+        }
+    }
+}
```
hamorillo, line 51: ⛏

```suggestion
}

@Composable
@Preview(showBackground = true)
private fun UploadNewAvatarSectionPreview() {
    MaterialTheme {
        UploadNewAvatarSection(
            onTakePictureClicked = {},
            onChooseFromGalleryClicked = {}
        )
    }
}

```

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModelTest.kt
```diff
@@ -142,6 +150,13 @@ class GravatarViewModelTest {
         coVerify { userRepository.selectAvatar(avatarId) }
     }
 
+    private fun initViewModel() {
```
hamorillo: ⛏️  Maybe we can move these methods to the end of the file instead of in the middle of the tests.

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModelTest.kt
```diff
@@ -153,4 +168,116 @@ class GravatarViewModelTest {
...
+        advanceUntilIdle()
+
+        mockkStatic("androidx.core.net.UriKt")
+        val file = mockk<File>()
+        val mockUri = mockk<Uri> {
+            every { toFile() } returns file
+        }
+        every { fileUtils.deleteFile(mockUri) } returns Unit
+
+        val newAvatar = createAvatars(5).last()
```
hamorillo: ⛏️  Not a big deal, but we're creating five avatars just to pick the last one, essentially to make it different. We can refactor the `createAvatars` method and extract the avatar creating to a method that can be invoked alone. 

Another option is to create more avatars in the Given and use all except one, which could be the `newAvatar`.

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/data/RealUserRepositoryTest.kt
```diff
@@ -342,6 +354,126 @@ class RealUserRepositoryTest {
         coVerify { profileService.updateProfileCatching(testToken, updateRequest) }
     }
 
+    @Test
+    fun `uploadAvatar should return success when user is logged in and services return success`() =
+        runTest {
+            // Given
+            val testFile = mockk<File>()
+            val profile = createTestProfile()
+            val uploadedAvatar = createTestAvatars().first()
```
hamorillo: ⛏️ Similar to a previous comment.

hamorillo (approved):
Works great! I left some ⛏️  comments.

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModelTest.kt
AdamGrzybkowski: yeah, for sure, I have to speak with Junie about file formatting.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/FileUtils.kt
AdamGrzybkowski: Agree, not sure either at this point. Those files could be module-specific, so I don't know if a shared module is the way to go.

PR #28, opened 2025-07-02
Open MediaPicker to select new avatar for upload

AdamGrzybkowski:
### Description

A follow-up with the `Photos` button click handler. It opens the Media Picker.

There's one addition in the presentation logic when the image is uploaded. If we try to upload the same file twice, the backend will not create a new avatar but rather move it to the first place, which means we have to filter out the old one from the list, otherwise the `LazyList` will crash as it will have duplicated keys.

### Testing Steps

1. Launch the app
2. Tap `Photo` button
3. Pick an image and crop it
4. Confirm it was uploaded
5. Repeat it with the same image
6. Confirm there's no crash or duplicated avatars

### DIFF
--- a/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarScreen.kt
+++ b/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarScreen.kt
@@ -6,6 +6,7 @@ import android.graphics.Color
 import android.net.Uri
 import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.ActivityResultLauncher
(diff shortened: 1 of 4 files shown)


hamorillo (commented):
Works as described! 

⚠️ I've found a corner case that can lead to weird behaviors. If, for some reason, I click twice on the `Photos` button, we'll open two mediapickers.

https://github.com/user-attachments/assets/f1f6395e-a118-4828-93b5-fc5cf6a4aec8

I've also tested it in a physical device, and it is "easy" to reproduce. I think we should take care of it.

AdamGrzybkowski:
Thanks @hamorillo - this should be fixed now.

hamorillo (approved):
:shipit:

PR #29, opened 2025-07-02
Gravatar tab collapsed header

hamorillo:
### Description

This PR implements a new composable for the collapsed version of the Gravatar Tab header. 

https://github.com/user-attachments/assets/b89cf1f6-ed53-4ae7-842c-e31b949b13b0

I've added the 'More Options' button to see how it looks with different Gravatars as a background. 

⚠️ **We still need to define what happens if there is no Gravatar selected.**

### Testing Steps

- Launch the app
- Switch between different Gravatars
- Verify that the header updates the avatars and background as expected


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
```diff
@@ -0,0 +1,101 @@
...
+                    .fillMaxWidth()
+                    .systemBarsPadding()
+            ) {
+                GravatarAvatar(
+                    state = ComponentState.Loaded(url),
+                    size = 44.dp,
+                    modifier = Modifier
+                        .clip(CircleShape)
+                        .align(Alignment.CenterVertically)
+                )
```
AdamGrzybkowski: Those small avatars seem to have a tiny 1px border with some kind of transparency gradient with a shadow as well. 

This may not be 100% accurate but something like this:
```kotlin 
                val colorStops = arrayOf(
                    0.0f to Color.Black.copy(0.3f),
                    0.6f to Color.Black.copy(alpha = 0f)
                )
                Box(
                    modifier = Modifier
                        .size(44.dp)
                        .clip(CircleShape)
                        .background(Brush.linearGradient(colorStops = colorStops))
                        .padding(1.dp)
                        .shadow(1.dp, CircleShape)
                ) {
                    GravatarAvatar(
                        state = ComponentState.Loaded(url),
                        size = 43.dp,
                        modifier = Modifier
                            .clip(CircleShape)
                    )
                }
```


| Design | Code above |
|---|---|
| Screenshot 2025-07-02 at 14 30 59 | Screenshot 2025-07-02 at 14 30 45 |

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
```diff
@@ -0,0 +1,101 @@
...
+                model = url,
+                contentDescription = null,
+                contentScale = ContentScale.Crop,
+                modifier = Modifier
+                    .matchParentSize()
+                    .blur(
+                        radius = 24.dp,
+                        edgeTreatment = BlurredEdgeTreatment.Rectangle,
+                    )
+            )
```
AdamGrzybkowski: 💡  The switch between the avatars is quite rough (when the image is not cached) - first the old avatar disappears and the header is empty, and then the new one appears, it's quite noticeable. Could we somehow make this transition nicer?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
```diff
@@ -0,0 +1,101 @@
...
+    avatarUrl?.let { url ->
+        Box(modifier.fillMaxWidth()) {
+            AsyncImage(
+                model = url,
+                contentDescription = null,
+                contentScale = ContentScale.Crop,
+                modifier = Modifier
+                    .matchParentSize()
+                    .blur(
+                        radius = 24.dp,
```
AdamGrzybkowski: ⛏️ I think the blur radius is higher in recent designs.

AdamGrzybkowski (commented):
Looks good, some design comments left 🎨

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
```diff
@@ -0,0 +1,101 @@
...
+import coil.compose.AsyncImage
+import com.gravatar.app.homeUi.R
+import com.gravatar.restapi.models.Avatar
+import com.gravatar.ui.components.ComponentState
+import java.net.URI
+import com.gravatar.ui.components.atomic.Avatar as GravatarAvatar
+
+@Composable
+fun GravatarHeader(
+    avatar: Avatar?,
```
AdamGrzybkowski, line 48: One more thing, more as an FYI: This will work on the GravatarTab, but on the others we might have to use the hash to get the current avatar, so maybe it's worth taking this into consideration already.

hamorillo: It appears that the header won't be shared between tabs (at least based on current designs). 

In this tab, we need to show the avatar switch as fast as possible, so using this approach allows us to use the cache (internally by Coil) when available. Using the hash will probably require dealing with the cacheRefresh / cacheBuster etc. 

We may need to address that in other tabs, but I would keep this approach for at least now.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
hamorillo: You have good eyes. Thanks for the code, I've used it, and that was very accurate.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
hamorillo: I agree. I've added a crossfade transition to the background, making the avatar switch smoother. There is still room for improvement, but it could be enough for now, as we'll need to revisit this with the expanded state. Wdty?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
```diff
@@ -55,21 +60,27 @@ fun GravatarHeader(
...
                     .systemBarsPadding()
             ) {
-                GravatarAvatar(
-                    state = ComponentState.Loaded(url),
-                    size = 44.dp,
-                    modifier = Modifier
-                        .clip(CircleShape)
-                        .align(Alignment.CenterVertically)
-                )
+                AvatarShadow(shape = CircleShape, size = 45.dp, modifier = Modifier.align(Alignment.CenterVertically)) {
```
AdamGrzybkowski: 💡 One more thing - It looks a bit weird when the image is being loaded from the network, the gradient with shadow is visible but nothing else, we might have to hide it or try to preserve the previous image. 

We may leave for now, but this is something we need to address for sure.  WDYT?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
AdamGrzybkowski: Looks like it might not be that complicated - [in case you want to add it or later](https://stackoverflow.com/questions/79230448/android-coil-loads-removes-old-image-to-early-when-loading-new-image).

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
AdamGrzybkowski: Screenshot 2025-07-03 at 08 40 13

hamorillo: 🤯  I’m going crazy with the shadow stuff. Adding the placeholder content is doable (which improves it a bit), but not at the moment when Coil, for some reason, removes everything. 

Trying to remove the shadow in that moment is not quite easy, or at least, I’m struggling with it. We need to use the AsyncImage callbacks for it.

Anyway, I'm not happy with the current solution, but to be totally honest, I think it's enough for the state of the project. The situation in your screenshot shouldn't occur right now; however, I must acknowledge that we still have room for improvement.

I think we need to explore the [preload of images](https://coil-kt.github.io/coil/faq/#how-do-i-preload-an-image).

Please give it a try, review the code, and let me know your opinion and ideas. Let me know if you think we should invest more time in this at the moment (in my opinion, we shouldn't).

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
```diff
@@ -0,0 +1,180 @@
...
+) {
+    val avatarUrl = avatar?.imageUrl?.toString() ?: ""
+
+    Box(modifier.fillMaxWidth()) {
+        GravatarAvatar(
+            avatarUrl,
+            modifier = Modifier
+                .matchParentSize()
+                .blur(radius = 40.dp, edgeTreatment = BlurredEdgeTreatment.Rectangle)
+        )
```
AdamGrzybkowski: Is blur the only effect we apply here? Should we have an additional overlay to make sure the white button will look correct on bright avatars?

AdamGrzybkowski (approved):
Looks much better now!

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/GravatarHeader.kt
hamorillo: One solution is to adjust the icon's alpha value. Reducing the alpha will make the icon more visible. 
I've slightly reduced the alpha.

PR #30, opened 2025-07-02
Retry failed avatar uploads

AdamGrzybkowski:
### Description

This PR adds each failed avatar upload to the UI state. When tapped an alert dialog is shown with extra information and an option to either retry or to remove the image. 

https://github.com/user-attachments/assets/601a38f4-7f1a-41ea-a264-533a0b7b523e

Keep in mind that the backend doesn't send proper error messages now, so you will likely see an empty dialog. 

### Testing Steps

1. You can either disable the internet, mock the response from the server or modify the aspect ratio when launch UCrop
2. Once the above is ready, launch app
3. Try to upload an image 
4. When the loading end, confirm you can see an avatar with warning overlay
5. Tap on it
6. Retry or remove, confirm works as expected


hamorillo (approved):
Looks good! 

Just a minor detail, not directly related to this as the majority of the code comes from the SDK, but when you have several images in error state and retries the last one, there is a weird jump to the first position:

https://github.com/user-attachments/assets/049179d6-80ca-496d-8f90-fd5f02c13aec

As I said, the SDK works in a similar way, but we can improve that in the future.

AdamGrzybkowski:
>Just a minor detail, not directly related to this as the majority of the code comes from the SDK, but when you have several images in error state and retries the last one, there is a weird jump to the first position:

True, I haven't changed it completely, but I did two small tweaks:
1. The items are now animated, so changing positions is a bit smoother.
2. Failed avatars are now first, and the uploading avatar is shown after. This means that when the avatar is uploaded after retrying it will stay in the same position.

PR #31, opened 2025-07-03
Avatar selection queue

AdamGrzybkowski:
### Description

This PR adds a queue for the avatar selection event to make sure only the latest one is executed and those in progress are dropped. Additionally, this will now skip making an API call if the user tries to select an already selected avatar. 

Here's ther recording showing three scenarios: 
1. Selecting already selected avatar (Nothing happens)
2. Continuously selecting new ones while API call is in progress  (Only the last one is selected)
3. Continuously selecting new ones while API call is in progress **and at the end selecting an already selected one** (API calls are dropped and the state restored)

https://github.com/user-attachments/assets/67982efb-8943-48f9-ad4a-b6a88e09f1cb

### Testing Steps

1. Launch the app
2. Try selecting the already selected avatar
3. Try selecting multiple ones quickly - confirm only the last one is selected
4. Try selecting multiple ones quickly and end with the selected one - confirm the progress bar is gone and the selected avatar stays selected without loading.

### DIFF
--- a/config/detekt/detekt.yml
+++ b/config/detekt/detekt.yml
@@ -44,7 +44,7 @@ style:
     excludeImports: []
   MaxLineLength:
     active: true
-    maxLineLength: 120
+    maxLineLength: 140
     excludePackageStatements: true
     excludeImportStatements: true
     excludeCommentStatements: false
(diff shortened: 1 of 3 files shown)


config/detekt/detekt.yml
```diff
@@ -44,7 +44,7 @@ style:
     excludeImports: []
   MaxLineLength:
     active: true
-    maxLineLength: 120
+    maxLineLength: 140
```
AdamGrzybkowski, line 47: `MaxLineLength` and `MaximumLineLength` were missaligned.

hamorillo (approved):
🚀

PR #32, opened 2025-07-04
Store user profile locally - PART 1

AdamGrzybkowski:
### Description

This is the first PR out of 2 for the local profile storage. 

This one creates a new `ProfileRepository` to handle the mediation logic between the remote and local Profile data sources. 

Temporarily, it holds the `Profile` property to serve as an in-memory local storage, but that will be handled by a proper storage framework in the next PR.

The idea here is that we will rely mostly on the locally cached profile and only update when requested. Such requests can come from different events: profile update on one of the screens, or we could also decide to call a refresh on each app startup. 

Additionally, I've introduced three use cases: Login/Logout/IsUserLoggedIn. The logic when logging in/out now requires more actions than just storing/clearing the user token, so we needed a place outside the `AuthRepository` to do that. This means the `AuthRepository` can now be internal.

### Testing Steps

1. Launch the app and make sure all works as before
2. You can verify the login-logout scenario - I've added a Logout action in the Gravatar tab.
3. An easy way to see a difference is when you launch the app and wait for the profile to load. Then switch to the Profile tab - it shows the data instantly.

### DIFF
--- a/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarScreen.kt
+++ b/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarScreen.kt
@@ -28,6 +28,7 @@ import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
@@ -45,19 +46,25 @@ import com.gravatar.app.homeUi.presentation.home.gravatar.components.GravatarHea
 import com.gravatar.app.homeUi.presentation.home.gravatar.components.UploadNewAvatarSection
 import com.gravatar.app.homeUi.presentation.home.gravatar.components.avatarSize
 import com.gravatar.app.homeUi.presentation.home.gravatar.components.avatarsGridSection
+import com.gravatar.app.usercomponent.domain.usecase.Logout
 import com.gravatar.restapi.models.Avatar
 import com.yalantis.ucrop.UCrop
 import com.yalantis.ucrop.UCropActivity
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import org.koin.androidx.compose.koinViewModel
+import org.koin.compose.koinInject
 import java.io.File
 import java.net.URI
+import kotlin.math.log
 
 @Composable
 internal fun GravatarScreen(
     viewModel: GravatarViewModel = koinViewModel()
 ) {
+    val logout = koinInject<Logout>()
+    val scope = rememberCoroutineScope()
     val uiState by viewModel.uiState.collectAsState()
     val lifecycle = LocalLifecycleOwner.current
     val context = LocalContext.current
@@ -123,6 +130,9 @@ internal fun GravatarScreen(
                 pickMedia.launch(PickVisualMediaRequest(ActivityResultContracts.PickVisualMedia.ImageOnly))
                 mediaPickerLaunched = true
             }
+        },
+        onMenuClick = {
+            scope.launch { logout() }
         }
     )
 }
@@ -131,6 +141,7 @@ internal fun GravatarScreen(
 @Composable
 internal fun GravatarScreen(
     uiState: GravatarUiState,
+    onMenuClick: () -> Unit = {},
     onTakePictureClicked: () -> Unit,
     onPickMediaClicked: () -> Unit,
     onEvent: (GravatarEvent) -> Unit = {},
@@ -150,6 +161,7 @@ internal fun GravatarScreen(
                 GravatarHeader(
                     uiState.avatars.firstOrNull { it.imageId == uiState.selectedAvatarId },
                     modifier = Modifier.fillMaxWidth(),
+                    onMenuClick = onMenuClick,
                 )
                 LazyVerticalGrid(
                     columns = GridCells.Adaptive(minSize = avatarSize),
--- a/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarScreen.kt
+++ b/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarScreen.kt
@@ -57,7 +57,6 @@ import org.koin.androidx.compose.koinViewModel
 import org.koin.compose.koinInject
 import java.io.File
 import java.net.URI
-import kotlin.math.log
 
 @Composable
 internal fun GravatarScreen(
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealProfileRepository.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealProfileRepository.kt
@@ -1,0 +1,69 @@
+package com.gravatar.app.usercomponent.data
+
+import com.gravatar.app.usercomponent.domain.repository.ProfileRepository
+import com.gravatar.restapi.models.Profile
+import com.gravatar.restapi.models.UpdateProfileRequest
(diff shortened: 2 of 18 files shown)


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarScreen.kt
```diff
@@ -150,6 +156,11 @@ internal fun GravatarScreen(
                 GravatarHeader(
                     uiState.avatars.firstOrNull { it.imageId == uiState.selectedAvatarId },
                     modifier = Modifier.fillMaxWidth(),
+                    onMenuClick = {
+                        scope.launch {
+                            logout()
+                        }
+                    }
```
AdamGrzybkowski: We have lost the logout action along the way - I've added it here for now.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
```diff
@@ -1,28 +1,28 @@
...
-import com.gravatar.services.ProfileService
 import com.gravatar.types.Hash
 import kotlinx.coroutines.flow.firstOrNull
 import java.io.File
 
 internal class RealUserRepository(
     private val avatarService: AvatarService,
-    private val profileService: ProfileService,
     private val tokenStorage: AuthTokenStorage,
+    private val profileRepository: ProfileRepository
```
AdamGrzybkowski, line 18: One question I have is whether we will need the `UserRepository` at some point. If we extract the `AvatarService` in the future, which will likely happen. `UserRepository` will only act as a pointer to other repositories - we could eliminate it in favor of use cases or use other repositories individually.  WDYT?

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/UserComponentModule.kt
```diff
@@ -1,19 +1,32 @@
...
+import com.gravatar.app.usercomponent.domain.usecase.LogoutUseCase
 import com.gravatar.services.AvatarService
 import com.gravatar.services.ProfileService
 import org.koin.core.module.dsl.bind
 import org.koin.core.module.dsl.factoryOf
+import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.module
 
 val userComponentModule = module {
+    singleOf(::RealProfileRepository) { bind<ProfileRepository>() }
```
AdamGrzybkowski, line 24: Single for now as it holds the profile property.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/LoginUseCase.kt
```diff
@@ -0,0 +1,22 @@
...
+internal class LoginUseCase(
+    val authRepository: AuthRepository,
+    val profileRepository: ProfileRepository,
+) : Login {
+
+    override suspend fun invoke(loginRequest: LoginRequest): Result<Unit> {
+        return authRepository.login(loginRequest)
+            .onSuccess {
+                profileRepository.refreshUserProfile()
+            }
```
AdamGrzybkowski, line 20: The idea is to refresh the profile, but not fail the login if the profile refresh fails.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealProfileRepository.kt
```diff
@@ -0,0 +1,69 @@
...
+            onSuccess = {
+                Result.success(Unit)
+            },
+            onFailure = { error ->
+                Result.failure(error)
+            }
+        )
+    }
+
+    override suspend fun get(): Result<Profile> {
```
AdamGrzybkowski, line 28: This might be changed to `Flow<Profile`, but we aren't really using it like that at this, so I've kept it like so.

hamorillo: 👍 Yeah, we can keep it for now, but for example, it would be nice to have a flow here when we implement the Share Tab, as we'll use some info from the profile that can be updated in a different tab. 

**Edited:** I left another comment referring to this.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/LoginUseCase.kt
AdamGrzybkowski: A note here: this should be invoked synchronously, but we observe the `token` for determining whether the user is logged in or not. That means the root navigator will open the HomeScreen before this finishes.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
hamorillo: When the Avatars repository is ready, I will drop the UserRepository. `User` is a virtual entity we've created because it was convenient, but when we have concrete repositories for `Profile` and `Avatar`, I don't think we need that anymore. We can use the repositories individually.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/LoginUseCase.kt
hamorillo: > That means the root navigator will open the HomeScreen before this finishes.

This could be a good reason to implement the `get` profile in the Profile as a flow as soon as possible. The `IsUserLoggedIn` can depend on the Profile instead of in the token, as we'll open the HomeScreen with the profile already loaded. 

With the current implementation, we are duplicating the `profile` request, as it's requested immediately after the login, but also immediately in the `GravatarTab`, as it needs to load the avatars.

image

We can address this in a separate PR, but we need to improve it, whether through my suggested approach or an alternative one.

hamorillo (commented):
👏 It works well, and I like the proposed architecture. I left some comments to discuss. 

Additionally, I think we can add some basic unit tests to the Use Cases (I know the current login is not complex, but they won't hurt and Junie should help with it quickly 🙂 ).

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/LoginUseCase.kt
AdamGrzybkowski: >The IsUserLoggedIn can depend on the Profile instead of in the token

I'm not sure about this idea. The `refreshProfile` result in the Login shouldn't block the user from being navigated to the Home screen. If we want to determine the user's logged-in state based on the cached profile, that's what we would have to do: fail the Login and force the user to repeat it.

AdamGrzybkowski: >This could be a good reason to implement the get profile in the Profile as a flow as soon as possible.

Also, I would prefer this to be done in a separate PR because this will require changes to the presentation logic. Not only in the way we consume the profile data, but also in resolving potential conflicts with already modified profile data on the profile tab.

AdamGrzybkowski:
>Additionally, I think we can add some basic unit tests to the Use Cases

Done, including one for ProfileRepository.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/LoginUseCase.kt
hamorillo: > If we want to determine the user's logged-in state based on the cached profile

Perhaps we should not determine the logged-in state based on the cached profile, as it is true that we are merging two different concepts. However, I'm not sure if we should navigate to the HomeScreen without a Profile, as the app will be entirely useless. We need the profile for the three tabs.

Anyway, we can discuss this outside this PR.

hamorillo (approved):
🚀  Thanks for adding the tests.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/LoginUseCase.kt
AdamGrzybkowski: I agree, I'm planning to tackle it next.

PR #33, opened 2025-07-04
Update ProfileHeader to use the blurred avatar as background

hamorillo:
### Description

This PR updates the profile header to align with the current design. The blurred user's avatar is used as a header background. It's worth noting that when the avatar is changed in the Gravatar tab, the update won't be reflected in the profile tab immediately. 

We are using the avatar URL from the profile, and with that data, the problem is not very straightforward to solve. However, as we plan to store the avatars and will know if the user switches the selected one, a possible solution could be to expose a flow that notifies of the current avatar chosen. We should continue the iteration to improve this header.

| Light Mode | Dark Mode |
|----------|-----------|
|image|image|

### Testing Steps

- Launch the app
- Switch to the profile tab and verify that the header looks as per the designs


AdamGrzybkowski (approved):
LGTM 👍 

> However, as we plan to store the avatars and will know if the user switches the selected one, a possible solution could be to expose a flow that notifies of the current avatar chosen. We should continue the iteration to improve this header.

Yeah, we will need a single place that will provide the latest avatar URL throughout the whole app.

PR #34, opened 2025-07-07
Store user profile locally - persistent local storage

AdamGrzybkowski:
### Description

This is a follow-up PR to https://github.com/Automattic/Gravatar-Android/pull/32. The `profile` property that was temporarily set in the `ProfileRepostiory` is now replaced with `ProfileDao`.

For now, we are only storing what we need (or those fields that are non-nullable. 

### Testing Steps

1. Launch the app and make sure all works as before
2. You can verify the login-logout scenario - I've added a Logout action in the Gravatar tab.
3. An easy way to see a difference is when you launch the app and wait for the profile to load. Then switch to the Profile tab - it shows the data instantly.


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/database/model/ProfileEntity.kt
```diff
@@ -0,0 +1,98 @@
...
+    @ColumnInfo(name = "company")
+    val company: String,
+    @ColumnInfo(name = "first_name")
+    val firstName: String? = null,
+    @ColumnInfo(name = "last_name")
+    val lastName: String? = null,
+    @ColumnInfo(name = "contact_cell_phone")
+    val contactCellPhone: String? = null,
+    @ColumnInfo(name = "contact_email")
+    val contactEmail: String? = null,
```
AdamGrzybkowski, line 44: `ContactInfo` is a separate object - I've put those fields here directly and flattened it rather than creating a 2nd table. Thoughts?

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/database/model/ProfileEntity.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    companion object {
+        /**
+         * Creates a ProfileEntity from a Profile model.
+         */
+        fun fromProfile(profile: Profile): ProfileEntity {
+            return ProfileEntity(
+                userId = profile.userId!!,
```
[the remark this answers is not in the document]
AdamGrzybkowski: This is intentional - `userId` is non-null for authenticated calls, and this is what we are using.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/database/UserDatabase.kt
```diff
@@ -0,0 +1,33 @@
+package com.gravatar.app.usercomponent.data.database
+
+import android.content.Context
+import androidx.room.Database
+import androidx.room.Room
+import androidx.room.RoomDatabase
+import com.gravatar.app.usercomponent.data.database.model.ProfileEntity
+
+@Database(entities = [ProfileEntity::class], version = 1, exportSchema = true)
```
hamorillo, line 9: We set the `exportSchema=true`, which I think is good, but we also need to [set the schema location and commit it to the repo.](https://developer.android.com/training/data-storage/room/migrating-db-versions#set_schema_location_using_room_gradle_plugin)

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/database/model/ProfileEntity.kt
hamorillo: I think that's fine.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/DatabaseModule.kt
```diff
@@ -0,0 +1,11 @@
+package com.gravatar.app.usercomponent.di
+
+import com.gravatar.app.usercomponent.data.database.ProfileDao
+import com.gravatar.app.usercomponent.data.database.UserDatabase
+import org.koin.android.ext.koin.androidContext
+import org.koin.dsl.module
+
+internal val databaseModule = module {
```
hamorillo, line 8: 💡⛏ Do you want to add a test with the verify for this module?

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/database/ProfileDao.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.gravatar.app.usercomponent.data.database.model.ProfileEntity
+
+@Dao
+interface ProfileDao {
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertProfile(profile: ProfileEntity)
+
+    @Query("SELECT * FROM user_profiles LIMIT 1")
+    suspend fun getProfile(): ProfileEntity?
```
hamorillo, line 16: 💡 Not necessary for this PR, but this can return a Flow:
```suggestion
    suspend fun getProfile(): Flow<ProfileEntity?>
```

hamorillo (commented):
Nice improvement! I left some comments.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/database/UserDatabase.kt
AdamGrzybkowski: Done

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/DatabaseModule.kt
AdamGrzybkowski: Done

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/database/ProfileDao.kt
AdamGrzybkowski: Yeah, I would keep this as part of the transition to collecting profiles in the presentation layer.

hamorillo (approved):
🚀  Thanks for the changes!

PR #35, opened 2025-07-08
Fix wrong recomposition when orientation changes

hamorillo:
### Description

When the orientation changed, we were losing the state. That was because we weren't saving the state in the RootNavigation.

<del>⚠️ This PR fixes that problem; however, it introduces a minor issue related to the back button. After an orientation change, the backButton is not handled by the nested NavGraph (`HomeScreen`) and it's received in the first NavGraph (`RootNavigation`). At the moment, it's not a big deal because it dismisses the app instead of moving the Gravatar Tab (in case you are not already on it). 
I would say that's something Compose is not doing well, and I haven't found a straightforward solution for it.  (If you have any ideas, please let me know)</del>

<del>Solving the wrong recomposition has higher priority than the back button, but of course, we'll need to find a proper solution. Wdty?</del>

Solved in [f21f1d7](https://github.com/Automattic/Gravatar-Android/pull/35/commits/f21f1d7ef31b4cef7536fab7d180d09afd28b2b3)

### Testing Steps

- Launch the app
- Navigate to the Profile Tab
- Rotate the device to force an orientation change
- Verify after the rotation that you are still on the Profile Tab (you can also verify that if you edit any fields, the states will be preserved)

### DIFF
--- a/app/src/main/java/com/gravatar/app/navigation/RootNavigation.kt
+++ b/app/src/main/java/com/gravatar/app/navigation/RootNavigation.kt
@@ -2,6 +2,7 @@ package com.gravatar.app.navigation
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
@@ -29,23 +30,9 @@ fun RootNavigation() {
         isUserLoggedIn()
             .collect { isLoggedIn ->
                 if (isLoggedIn) {
-                    navController.navigate(HomeDest) {
-                        popUpTo(SplashDest) {
-                            inclusive = true
-                            saveState = true
-                        }
-                        launchSingleTop = true
-                        restoreState = true
-                    }
+                    navController.navigateSavingState(HomeDest)
                 } else {
-                    navController.navigate(LoginDest) {
-                        popUpTo(SplashDest) {
-                            inclusive = true
-                            saveState = true
-                        }
-                        launchSingleTop = true
-                        restoreState = true
-                    }
+                    navController.navigateSavingState(LoginDest)
                 }
             }
     }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/gravatar/app/navigation/RootNavigation.kt
```diff
@@ -50,3 +47,14 @@ fun RootNavigation() {
         }
     }
 }
+
+private fun NavHostController.navigateSavingState(destination: Any) {
+    navigate(destination) {
+        popUpTo(SplashDest) {
```
AdamGrzybkowski: Just wanted to note, this is not correct, and we have GRA-450 for it. During the login/logout flow, we should popUp to a different destination. 

Not part of this PR, but since the code is being modified, I figured it's worth mentioning.

AdamGrzybkowski (commented):
>After an orientation change, the backButton is not handled by the nested NavGraph (HomeScreen) and it's received in the first NavGraph (RootNavigation).

Definitely not ideal, if possible I would suggest to try and fix it as we will have to come back to it anyway.

app/src/main/java/com/gravatar/app/navigation/RootNavigation.kt
hamorillo: I've just pushed a commit that should resolve GRA-450 as well.

hamorillo:
The back issue has also been resolved in [f21f1d7](https://github.com/Automattic/Gravatar-Android/pull/35/commits/f21f1d7ef31b4cef7536fab7d180d09afd28b2b3)

AdamGrzybkowski (approved):
Great work! The back button issue is fixed 🎉 

BTW. I noticed that when you have unsaved changes in the Profile tab and press the back button, we open the avatar tab and the changes are lost on the profile. When you navigate to the avatar via the bottom tab click, the changes are restored. Not related to this PR, but something that we might want to change.

hamorillo:
> BTW. I noticed that when you have unsaved changes in the Profile tab and press the back button, we open the avatar tab and the changes are lost on the profile. When you navigate to the avatar via the bottom tab click, the changes are restored. Not related to this PR, but something that we might want to change.

Good catch! Let's work on that in a different PR 🙂

PR #36, opened 2025-07-08
UserSession persistence to keep and emit user logged in state

AdamGrzybkowski:
### Description

This PR introduced the `UserSessionPersistence` to keep full control of the user session state and be able to emit the actual logged in/out state when the app state is ready. 

Until now, we relied fully on the token storage, but this had a problem. We had to save it and perform certain operations after, but without navigating the user to the Home screen. This gives us the same reactive approach but with a full control over when the state changes. 

### Testing Steps

1. Test log in, log out flows. 
2. Test a fresh app open (after killing it) in each state to make sure the proper screen is shown at the start.

### DIFF
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/AuthTokenStorage.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/AuthTokenStorage.kt
@@ -1,19 +1,19 @@
 package com.gravatar.app.usercomponent.data
 
 import androidx.datastore.core.DataStore
+import androidx.datastore.core.IOException
 import androidx.datastore.preferences.core.Preferences
 import androidx.datastore.preferences.core.edit
 import androidx.datastore.preferences.core.stringPreferencesKey
 import com.gravatar.app.foundations.DispatcherProvider
 import com.gravatar.app.usercomponent.di.UserPrefs
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.flowOn
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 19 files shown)


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/AuthTokenStorage.kt
```diff
@@ -1,19 +1,18 @@
...
-import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.catch
-import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.withContext
 
 internal interface AuthTokenStorage {
-    fun get(): Flow<String?>
+    suspend fun get(): String?
```
AdamGrzybkowski, line 16: We no longer need the Flow here.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/InMemoryUserSessionPersistence.kt
```diff
@@ -0,0 +1,40 @@
...
+    authRepository: AuthRepository,
+    applicationScope: CoroutineScope,
+    dispatcherProvider: DispatcherProvider,
+) : UserSessionPersistence {
+
+    private val _state: MutableSharedFlow<UserSession> = MutableSharedFlow(replay = 1)
+    override val state: Flow<UserSession> = _state
+
+    init {
+        applicationScope.launch(dispatcherProvider.io) {
```
AdamGrzybkowski, line 21: In order to initialize the state, we have to read from the DataStore, which is a suspending operation, so we need a scope. I've created an Application scope for that.

hamorillo (approved):
Great improvement! 🚀

PR #37, opened 2025-07-08
GetAvatarUrl use case

AdamGrzybkowski:
### Description

This PR introduced the `GetAvatarUrl` use case that acts as a single point for getting the current avatar URL. 

The URL is built based on the currently stored Profile and the cache buster value. Whenever a new avatar is selected, the cache-buster param is updated and a new URL is emitted. 

https://github.com/user-attachments/assets/7946bc26-3282-4675-bc95-1a6b240315d2

### Testing Steps

1. Launch the app
2. Log in
3. Confirm you can see the avatars in the header
4. Go to profile tab, confirm the header looks correct
5. Go to avatar tab, select a new avatar
6. Confirm the header was updated
7. Go to Profile tab
8. Confirm the same avatar is visible
9. You can also test this by killing the app or opening the app without network connection
10. As long as the image was shown before, it should be visible without waiting for any network requests


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/GetAvatarUrl.kt
```diff
@@ -0,0 +1,32 @@
...
+internal class GetAvatarUrlUseCase(
+    private val profileRepository: ProfileRepository,
+    private val avatarCacheBusterStorage: AvatarCacheBusterStorage,
+) : GetAvatarUrl {
+
+    override fun invoke(): Flow<URL?> {
+        return avatarCacheBusterStorage.getAvatarCacheBuster()
+            .map { cacheBuster ->
+                val hash: String? = profileRepository.get()
+                    .getOrNull()?.hash
```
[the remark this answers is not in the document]
AdamGrzybkowski: `profileRepository.get()` will be modified and return flow later on, I would keep it for now.

homeUi/src/test/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModelTest.kt
```diff
@@ -34,9 +41,19 @@ class GravatarViewModelTest {
     var coroutineTestRule = CoroutineTestRule(testDispatcher)
 
     private val userRepository: UserRepository = mockk()
+    private val getAvatarUrl: GetAvatarUrl = mockk()
```
hamorillo: 💡 I wonder if we should try to use a custom implementation of the interfaces when possible and practical, rather than mocking them. For example, in this case, the change will be very simple:
```suggestion
    private val getAvatarUrl: GetAvatarUrl = object : GetAvatarUrl {
        override fun invoke() = avatarUrlFlow
    }
```
And we'll be able to remove the `setup` method.

userComponent/src/test/kotlin/com/gravatar/app/usercomponent/domain/usecase/GetAvatarUrlUseCaseTest.kt
```diff
@@ -0,0 +1,121 @@
...
+class GetAvatarUrlUseCaseTest {
+
+    private val testDispatcher = StandardTestDispatcher()
+
+    @get:Rule
+    var coroutineTestRule = CoroutineTestRule(testDispatcher)
+
+    private lateinit var getAvatarUrlUseCase: GetAvatarUrlUseCase
+    private val profileRepository = mockk<ProfileRepository>()
+    private val avatarCacheBusterStorage = mockk<AvatarCacheBusterStorage>()
```
hamorillo: 💡 Similar example as in a previous comment:

```suggestion
    private val avatarCacheBusterStorage = object : AvatarCacheBusterStorage {
        override fun getAvatarCacheBuster() = cacheBusterFlow
        override suspend fun saveAvatarCacheBuster(cacheBuster: String) {
            cacheBusterFlow.value = cacheBuster
        }
    }
```

AdamGrzybkowski: I did that, but only for this use case. I'm not sure I like this when the `saveAvatarCacheBuster` is used. I can't verify it was called and the test order with `cacheBusterFlow` collection is harder to follow.

hamorillo (approved):
Looks good! 🚀

PR #40, opened 2025-07-09
Observe profile via Flow

AdamGrzybkowski:
### Description

Refactoring the code to observe the locally stored Profile rather than pulling it when needed. 

This also closes GRA-451. The profile is refreshed in the `ProfileViewModel` init.

I haven't updated the presentation logic in this PR, but we should eventually do this:
- Show an error state when the profile == null and there's no loading - this means we need to refresh the profile as it's not available.
- Show some kind of message that the user is possibly seeing stale data when the refreshProfile fails.
- [Maybe] Show a warning when collecting the profile with some unsaved changes.

The last I've kept as maybe, because the current logic simply sets the new profile in the state, but the edited fields are not cleared - this means that the user will see the update and keep their changes. 

### Testing Steps

1. Generally, test all the flows. There's still a code that will try to fetch the profile once in the `UserRespository` when not available to get the hash, so if the profile fetch fails during login, we should recover. To test thi,s either mock the refresh during login, or remove the logic from the Login use case.
2. Test collecting a new profile with unsaved changes - I did that by changing the event sent on the save button click and modifying the profile on web.

### DIFF
--- a/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileUiState.kt
+++ b/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileUiState.kt
@@ -14,6 +14,8 @@ internal data class ProfileUiState(
 
     val originalAboutFields: Set<AboutEditorField> = profile?.aboutFields() ?: emptySet()
 
+    val showLoading = isLoading && profile == null
+
     val aboutFields: Set<AboutEditorField> = originalAboutFields.map { field ->
         val editedValue = editedAboutFields[field.type]
         if (editedValue != null) {
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
@@ -9,6 +9,8 @@ import com.gravatar.services.AvatarService
 import com.gravatar.services.ErrorType
 import com.gravatar.services.GravatarResult
 import com.gravatar.types.Hash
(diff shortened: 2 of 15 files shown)


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileUiState.kt
```diff
@@ -14,6 +14,8 @@ internal data class ProfileUiState(
 
     val originalAboutFields: Set<AboutEditorField> = profile?.aboutFields() ?: emptySet()
 
+    val showLoading = isLoading && profile == null
```
AdamGrzybkowski, line 17: We only want to show the loading when the profile is null; otherwise, the loading should happen in the background.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealUserRepository.kt
```diff
@@ -102,4 +105,15 @@ internal class RealUserRepository(
...
+    // Retrieves the profile, either from local storage or by refreshing it if not available
+    private suspend fun getOrFetchProfile(): Profile? {
+        val localProfile = profileRepository.get().firstOrNull()
+        return localProfile
+            ?: profileRepository.refreshUserProfile()
+                .fold(
+                    onSuccess = { profileRepository.get().firstOrNull() },
+                    onFailure = { null }
+                )
+    }
```
AdamGrzybkowski, line 118: This is necessary to attempt recovery if the profile fetch fails during login.

hamorillo (approved):
Works great! 🚀

PR #42, opened 2025-07-10
Error state for null profile in the ProfileScreen

AdamGrzybkowski:
### Description

Error state for when the profile is null. We read the profile from the local DB, so this would only be shown if the refresh profile fails after login and on each subsequent refresh. That means, in most scenarios, it won't be shown, but we need to handle that case.

Additionally, I've moved the ProfileHeader higher up, and it will be displayed while loading is in progress. 

Screenshot_20250710_102231

### Testing Steps

1. Launch the app, log out
2. Monitor the network and modify the profile refresh response to fail
3. Log in
4. Go to the profile tab and confirm that the error state is visible
5. Refresh the profile without modifying the response
6. Confirm the profile fields are visible


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
```diff
@@ -43,41 +48,69 @@ internal fun ProfileScreen(uiState: ProfileUiState, onEvent: (ProfileEvent) -> U
...
+                uiState.showLoading -> {
+                    Box(modifier = Modifier.fillMaxSize()) {
+                        CircularProgressIndicator(
+                            modifier = Modifier.align(Alignment.Center)
+                        )
+                    }
+                }
+
+                uiState.profile == null -> {
+                    Column(
```
hamorillo, line 72: ⛏ We can extract this to a Composable to reduce the code inside `ProfileScreen`.

hamorillo:
The code looks good however, I have some doubts about the approach:

- I'm not sure how you get the state from the screenshot in the PR description. The avatar in the header requires the `avatarUrl`, which will be `null` if we don't have a profile saved in the DB.

So the user won't see the header, just the empty space. It doesn't hurt, but I suppose that's not what you're expecting. 

image

- If we don't have a profile saved, the Gravatar tab will also be completely empty and the actions the user sees won't work. It's true that both actions, `Camera` and `Photos`, will perform a profile request, so that's a good point because if it works at that point, the app will be recovered. 

| Empty Gravatar Tab | After uploading an avatar (Profile still failing) |
|----------|-----------|
|image|image|

- When we implement the Share tab, we will suffer the same problems if there is no profile already stored.

In my opinion, we need to handle this case differently. The profile is basic for the whole app, and I would say we should show the `HomeScreen` if we don't have a Profile stored in the DB.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
AdamGrzybkowski: Yes, I'm doing it in the next PR already!

PR #43, opened 2025-07-10
Add GravatarSnackbarHost in the HomeScreen scaffold

hamorillo:
### Description

We need to inform about different events through a Snackbard. This PR adds the `SnackbarHost` to the `Scaffold` used in the HomeScreen so it can be used in any of the tabs. 

To test the implementation, we are implementing the confirmations for when a new avatar is selected. Both error and success.

https://github.com/user-attachments/assets/62acd075-0fce-4800-805c-398dfe6fd9b7

### Testing Steps

- Switch between avatars and verify you can see the confirmation snackbard. 
- Turn off the network connection, try to switch between avatars, and verify you can see the error snackbar.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/Snackbar.kt
```diff
@@ -0,0 +1,78 @@
...
+        withDismissAction = withDismissAction,
+        actionLabel = actionLabel,
+        duration = duration,
+        snackbarType = snackbarType,
+    )
+
+    return showSnackbar(visuals)
+}
+
+internal val SnackbarType.containerColor: Color
```
hamorillo, line 63: I kept the SDK colors, as we are still discussing the styling. This should be easy to refactor.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
```diff
@@ -19,23 +21,26 @@ import androidx.navigation.compose.currentBackStackEntryAsState
...
-    content: @Composable (NavHostController) -> Unit
+    content: @Composable (NavHostController, SnackbarHostState) -> Unit
 ) {
     val navController = rememberNavController()
+    val snackbarHostState = remember { SnackbarHostState() }
     val backStackEntry = navController.currentBackStackEntryAsState()
 
     Scaffold(
         contentWindowInsets = WindowInsets(0, 0, 0, 0),
+        snackbarHost = { GravatarSnackbarHost(hostState = snackbarHostState) },
```
AdamGrzybkowski, line 43: 💡 This might seem convenient now, but it could lead to problems when we have to implement some avoidance with other UI elements on each tab. I would consider changing this layout to:
1. Simple Column here in the Home screen to host the Content and the NavBar
2. Using Scafold on each tab to the SnackbarHost and the TopAppBar

The second part might be handy for the expandable headers with the provided scrolling behaviors, although I haven't explored it enough to know if we can have custom TopAppBars.

AdamGrzybkowski: Also, the SnackHost can only have one SnackBar shown. If we have a snack that's not auto-dismissible on one tab, then other tabs would override it.

AdamGrzybkowski (commented):
The code works, but I've left a comment about the structure and to consider an alternative. Let me know what you think!

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/HomeScreen.kt
hamorillo: Thank you for providing a different perspective!

I'm not sure if I see more benefits to the suggested approach. 

> If we have a snack that's not auto-dismissible on one tab, then other tabs would override it.

I don't think that's true, the snackbars will pile up until the previous one is dismissed (I've modified the error one to not auto-dismiss):

https://github.com/user-attachments/assets/38797928-a608-40c9-a634-69b2e6a63de4

Also, in the designs, it mentioned that we should think of this "Toast" as something ephemeral, so, in general, we shouldn't use them in that way.
_____

Having a different SnackbarHost in each tab can lead the user to miss some information. For example, I'm in the Gravatar tab, I want to change my avatar, but the request is taking some time. If I switch to a different tab and the request fails, I would never know that. With the current approach, you still will see the snackbar. 

----
> it could lead to problems when we have to implement some avoidance with other UI elements on each tab

🤔 Any idea or situation in mind?

AdamGrzybkowski: >I don't think that's true, the snackbars will pile up until the previous one is dismissed (I've modified the error one to not auto-dismiss):

I got it the other way around. That is even worse, I think - because it will block the user from seeing the snack on a different tab.

>Having a different SnackbarHost in each tab can lead the user to miss some information. For example, I'm in the Gravatar tab, I want to change my avatar, but the request is taking some time. If I switch to a different tab and the request fails, I would never know that. With the current approach, you still will see the snackbar.

I'm not convinced that showing snackbars that are out of context (from a different tab) makes sense. There's a certain context on the screen, and the information inside the Snack is related to it. When we suddently show a sanckbar with an error from a different tab, it could be confusing. What about actions? If tapped, would we navigate to the tab that posted the Snack? 

 I checked a few Google apps with bottom bars and couldn't find one with such behavior. 
 
 ---
 
 >🤔 Any idea or situation in mind?
 
 With the current design, only the `Share` tab has a button that will be covered by the Snack. Not sure we will need snacks there.

hamorillo: I've been reading about the [Material3 recommendation for snackbars](https://m3.material.io/components/snackbar/guidelines).

They mentioned some topics we are discussing here, for example, about showing [several snackbars at the same time](https://m3.material.io/components/snackbar/guidelines#84cd5bb6-354f-4fd0-bdd9-daf9684ba4bf):

Unfortunately, I didn't see any recommendation about keeping the snackbar between tab switches. I've tried a couple of Google apps too and found different behaviours.

Gmail dismisses the snackbar of email deleted when switching to the Meet tab. However, YouTube preserves the snackbar between tabs:

https://github.com/user-attachments/assets/34051963-f888-4ad2-a7bc-57e8f7a5c2e1

I have to recognize that I have more doubts now, but I'm still not sold on either of them.

One benefit of the Scaffold with the SnackbarHost is that the Scaffold handles part of the presentation logic of the Snackbar. For example, the Snackbar is shown at the top of the BottomBar thanks to the Scaffold. If not, we'll need to handle this kind of issue in each tab.

AdamGrzybkowski: >For example, the Snackbar is shown at the top of the BottomBar thanks to the Scaffold. If not, we'll need to handle this kind of issue in each tab.


Tabs are already above the bottom nav, so that shouldn't be an issue. Anyway, let's keep it the way it is now. We can reevaluate later.

AdamGrzybkowski (approved):
:shipit:

PR #44, opened 2025-07-10
Error state for avatar loading

AdamGrzybkowski:
### Description

Following the same pattern as in the https://github.com/Automattic/Gravatar-Android/pull/42.

The UI is extracted to a shared composable and used on both screens with different strings.

Screenshot_20250710_124020

### Testing Steps

1. Mock the avatar request to fail
2. Launch the app 
3. Confirm you see the error state
4. Mock the avatar request to fail
5. Pull to refresh
6. Confirm you see the loading and after that the error state again
7. Tap retry (not mocking now)
8. Confirm the avatars are shown after loading is gone


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModel.kt
```diff
@@ -197,7 +198,7 @@ internal class GravatarViewModel(
             _uiState.update { currentState ->
                 currentState.copy(
                     isRefreshing = isRefreshing,
-                    isLoading = !isRefreshing
+                    isLoading = !isRefreshing || currentState.avatars == null,
```
AdamGrzybkowski, line 201: I think we should show the loading after pull to refresh in case the avatars are null. WDYT?

hamorillo: I agree! 

In fact, maybe in that case, when the avatars are null, we should hide the loading from the pull-to-refresh.

image

hamorillo (approved):
Looks very nice!

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModel.kt
AdamGrzybkowski: The pull to refresh freezes when you try to skip it. I will merge it for now.

PR #46, opened 2025-07-11
Fix uploadAvatar when no avatar is selected

hamorillo:
### Description

When a new avatar is uploaded, we need to check if that avatar has been selected and update our persistence if necessary.

https://github.com/user-attachments/assets/351d6e21-44f8-4cdc-93ad-9889b10850a8

Previously to this PR, we weren't reflecting the real status after uploading a new avatar.

### Testing Steps

- In the Gravatar Tab, delete your currently selected avatar
- Upload a new avatar with either of the options (Camera or Photos)
- Verify the new avatar is shown in the headers and as the selected avatar
- Upload a different avatar with either of the options (Camera or Photos)
- Verify there is no change in the selected avatar


AdamGrzybkowski (approved):
:shipit:

PR #39, opened 2025-07-09
Download avatar

hamorillo:
### Description

This PR adds the option to download the avatars. It uses the DownloadManager as we did in the SDK.

When the device API is < 29, the app should check and ask for permission to write to the external storage.

**Note:** Error handling is not implemented in this PR. We need to notify the UI through the ViewModel about the different errors and inform the user.

**API < 29 (In this case 28)**

https://github.com/user-attachments/assets/f21647dd-086a-468f-b756-17fa61aaf780

### Testing Steps

Please test the PR on a device with API < 29 and on a device with API >= 29.

**API < 29**
- Verify that the first time you try to download an avatar, you are asked for permission.
- Play with that flow: Deny the permission, verify the following dialog, and accept the permission. Verify that you are redirected to the app settings...
- Verify the avatar is finally downloaded to the device

**API >= 29**
- Verify the download flow works as expected.

PR #47, opened 2025-07-11
Add missing notification Snackbars

hamorillo:
### Description

This is a follow-up to #43. Now we are adding the missing snackbars for the following actions:

- Profile update (success-failure)
- Avatar update (success-failure)
- Delete avatar (success-failure)
- Download Avatar (started-download manager not available)

https://github.com/user-attachments/assets/973daaa5-4627-42cc-b728-37a1e01deca1

**Note:** We are using the same messages as the SDK. 

### Testing Steps

- For each of the mentioned flows, verify that you can see the corresponding toast. For error cases, you can disable the device's network or block the request.

### DIFF
--- a/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModel.kt
+++ b/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModel.kt
@@ -292,7 +292,7 @@ internal class GravatarViewModel(
                     is GravatarResult.Failure -> {
                         when (result.error) {
                             DownloadManagerError.DOWNLOAD_MANAGER_NOT_AVAILABLE -> {
-                                // Notify the UI that the download manager is not available - We should update tests to handle this case
+                                _actions.send(GravatarAction.DownloadManagerNotAvailable)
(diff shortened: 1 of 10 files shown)


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/GravatarViewModel.kt
```diff
@@ -292,7 +292,7 @@ internal class GravatarViewModel(
                     is GravatarResult.Failure -> {
                         when (result.error) {
                             DownloadManagerError.DOWNLOAD_MANAGER_NOT_AVAILABLE -> {
-                                // Notify the UI that the download manager is not available - We should update tests to handle this case
+                                _actions.send(GravatarAction.DownloadManagerNotAvailable)
                             }
 
                             DownloadManagerError.DOWNLOAD_MANAGER_DISABLED -> {
```
hamorillo, line 298: ⚠️ I realize that I forgot to implement this case in #39, which is not solvable with a toast. I've created a ticket and will open a PR soon.

AdamGrzybkowski (approved):
:shipit:

PR #48, opened 2025-07-11
Loading profile failure state on Login Screen

AdamGrzybkowski:
### Description

This updates the login flow with the new requirement of having the profile successfully fetched before showing the home screen.

The main changes:
- LoginUseCase now returns different LoginResults and fails if the profile refresh fails
- AuthRepository no longer does operations on `tokenStorage`
- UserSessionPersistence uses profileRepository now to init the UserSession state
- ~The Login button was moved slightly up to better align with the error state (not sure about this, especially in landscape mode)~ I have reverted this change for now.

I also need to show the Snackbar here, so I extracted `GravatarSnackbarHost` to the shared `:design` module.

| Initial state | OAuth denied | Load profile fail |
|---|---|---|
|  login_initial | login_oauth_denied | login_load_profile_fail |


loginUi/screenshotTests/roborazzi/com.gravatar.app.loginUi.presentation.login.LoginScreenTest.loginScreenLandscape.png
AdamGrzybkowski, line 1: This doesn't look good...

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginScreen.kt
```diff
@@ -55,89 +64,134 @@ fun LoginScreen() {
...
-                        LoginAction.ShowError -> {
-                            Toast.makeText(context, "Login error", Toast.LENGTH_SHORT)
-                                .show()
+                        is LoginAction.ShowLoginError -> {
+                            scope.launch {
+                                snackbarHostState.showLoginErrorSnack(context)
+                            }
                         }
+
+                        LoginAction.StartOAuth -> oAuthLauncher.launch(Unit)
```
AdamGrzybkowski, line 87: FYI: I have an issue with this - it stops working when changing orientations...

AdamGrzybkowski: This was fixed in https://github.com/Automattic/Gravatar-Android/pull/48/commits/46b3282e6564b40f2e36de9b97322ba1521e281e

loginUi/screenshotTests/roborazzi/com.gravatar.app.loginUi.presentation.login.LoginScreenTest.loginScreenLandscape.png
AdamGrzybkowski: Reverted that change

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealAuthRepository.kt
```diff
@@ -1,30 +1,22 @@
...
-            clientId = loginRequest.clientId
+            code = oAuthRequest.code,
+            clientSecret = oAuthRequest.clientSecret,
+            redirectUri = oAuthRequest.redirectUri,
+            clientId = oAuthRequest.clientId
         ).fold(
             onSuccess = { token ->
-                userStorage.saveToken(token)
-                Result.success(Unit)
+                Result.success(token)
```
hamorillo, line 19: ❓ Why are we not saving the `token` anymore here, and we moved it to the UseCase?

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginUiState.kt
```diff
@@ -0,0 +1,17 @@
...
+    val error: LoginError? = null,
+)
+
+internal sealed class LoginError {
+    data object AuthorizationDenied : LoginError()
+    data class ProfileLoadFailure(
+        val reason: Reason,
+    ) : LoginError() {
+        enum class Reason {
+            GENERIC_ERROR,
```
hamorillo, line 14: ❓ Do we plan to expand this enum in the short term?

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginViewModel.kt
```diff
@@ -29,36 +31,81 @@ internal class LoginViewModel(
...
+    }
+
+    private fun restartLogin() {
+        viewModelScope.launch {
+            _uiState.update {
+                it.copy(
+                    error = null
+                )
+            }
+            sendAction(LoginAction.StartOAuth)
```
hamorillo, line 47: 🤔 _Just thinking aloud._ I know this saves a click from the user; however, it feels a bit weird because I see the navigation, and immediately the browser is open. I guess that's what the user would expect, but I feel something strange happened.

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginScreen.kt
hamorillo: 😕 I don't deeply understand what was happening here. However, I've tested with and without the commit changes, and it resolves the issue, that's for sure.

hamorillo (approved):
Works well! 🚀  Left some questions.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealAuthRepository.kt
AdamGrzybkowski: Good question - I had to read tokenStorage directly so it felt weird to have it there and in the usecase. I could either change the auth repo to return local/remote, or move it back there and expose another function to get the stored token. Do you have a preference?

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginUiState.kt
AdamGrzybkowski: I hope so, at least one that will say it's due to the network connection.

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginViewModel.kt
AdamGrzybkowski: hmm I tried it without first and it was weird to have to click again, but I don't have a strong opinion about this. @etoledom Have you thought about how the `try with another account` button should behave?

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/RealAuthRepository.kt
hamorillo: I would say that TokenStorage should be only used through the repositories and not in the UseCases directly, so exposing a function in `AuthRepository` sounds reasonable.

The `fetchToken` will save the token, and we can have a different function to expose the token.

AdamGrzybkowski: Done

PR #49, opened 2025-07-14
Empty state for avatars grid

hamorillo:
### Description

This PR adds the "empty state" for the avatars grid. When the user doesn't have any avatar already uploaded, they'll see this state.

image

It's just a very basic state, we'll iterate over it.

### Testing Steps

- Launch the app and log in using an account with no avatar (or delete all avatars from one of your accounts).
- Verify you can see the label "Your avatars will show up here"


AdamGrzybkowski (approved):
:shipit:

PR #51, opened 2025-07-14
Profile Screen UI update

AdamGrzybkowski:
### Description

This PR:
- updates the input field styling and paddings to match Figma
- adds `edited/focused` states for the input field
- adds pull to refresh (enabled only if there are no edits done)

**Edited/focused states**
| Light | Dark |
|---|---|
| profile_ui_ligth | profile_ui_dark |

Recording:
https://github.com/user-attachments/assets/de79e9e5-efad-4c0b-8bff-b7793d2c9460

### Testing Steps

1. Open the profile tab and verify the new layout
2. Pull to refresh
3. Confirm it works
4. Tap on an input field
5. Confirm the blue border is visible
6. Edit the field
7. Focus on another one
8. Confirm that the previous has a slight blue background now
9. Try to pull to refresh
10. Confirm it didn't work


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
```diff
@@ -115,6 +125,36 @@ internal fun ProfileScreen(uiState: ProfileUiState, onEvent: (ProfileEvent) -> U
...
 
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun PullToRefreshBox(
+    enabled: Boolean,
+    isRefreshing: Boolean,
+    onRefresh: () -> Unit,
+    modifier: Modifier = Modifier,
+    content: @Composable BoxScope.() -> Unit
+) {
```
AdamGrzybkowski, line 136: The default component doesn't have an `enabled` property, so I had to create this.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/about/AboutEditField.kt
```diff
@@ -26,35 +33,53 @@ internal fun AboutEditField(
     modifier: Modifier = Modifier,
     maxLines: Int = 1,
     description: String? = null,
+    edited: Boolean = false,
 ) {
+    val inputFieldShape = RoundedCornerShape(8.dp)
+    var isFocused by remember { mutableStateOf(false) }
```
AdamGrzybkowski, line 39: Kept here intentionally. When we change orientation, the soft keyboard and the focus are gone, so I didn't want to have this state in the VM.

hamorillo (approved):
👨‍🎨  🚀  Works as expected!

PR #52, opened 2025-07-14
Handle unauthorized errors - 401

hamorillo:
### Description

This PR handles unauthorized errors from the server. When a 401 error is received from the server, the interceptor will perform a logout. 

### Testing Steps

- Launch the app connected to a network analyzer.
- Log in
- Override responses from the Avatar and/or Profile services to return a 401
- Verify that a logout is performed and you are redirected to the login screen

### DIFF
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
@@ -1,0 +1,34 @@
+package com.gravatar.app.usercomponent.data.interceptors
+
+import com.gravatar.app.foundations.DispatcherProvider
+import com.gravatar.app.usercomponent.domain.usecase.Logout
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import okhttp3.Interceptor
+import okhttp3.Response
+
+/**
+ * Interceptor to analyze responses for 401 errors.
+ *
+ * The logout is injected as a lazy dependency to avoid a circular dependency.
+ * ProfileService -> UnauthorizeInterceptor → LogoutUseCase → RealProfileRepository → ProfileService
+ *
+ */
+class UnauthorizeInterceptor(
+    private val applicationScope: CoroutineScope,
+    private val dispatcherProvider: DispatcherProvider,
+    private val logout: Lazy<Logout>
+) : Interceptor {
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        val response = chain.proceed(request)
+
+        if (response.code == HttpResponseCode.UNAUTHORIZED) {
+            applicationScope.launch(dispatcherProvider.io) {
+                logout.value.invoke()
+            }
+        }
+        return response
+    }
+}
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
@@ -1,6 +1,5 @@
 package com.gravatar.app.usercomponent.data.interceptors
 
-import com.gravatar.app.foundations.DispatcherProvider
 import com.gravatar.app.usercomponent.domain.usecase.Logout
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
@@ -16,7 +15,6 @@ import okhttp3.Response
  */
 class UnauthorizeInterceptor(
     private val applicationScope: CoroutineScope,
-    private val dispatcherProvider: DispatcherProvider,
     private val logout: Lazy<Logout>
 ) : Interceptor {
 
@@ -25,7 +23,7 @@ class UnauthorizeInterceptor(
         val response = chain.proceed(request)
 
         if (response.code == HttpResponseCode.UNAUTHORIZED) {
-            applicationScope.launch(dispatcherProvider.io) {
+            applicationScope.launch {
                 logout.value.invoke()
             }
         }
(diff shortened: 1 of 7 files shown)


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+ * Interceptor to analyze responses for 401 errors.
+ *
+ * The logout is injected as a lazy dependency to avoid a circular dependency.
+ * ProfileService -> UnauthorizeInterceptor → LogoutUseCase → RealProfileRepository → ProfileService
+ *
+ */
+class UnauthorizeInterceptor(
+    private val applicationScope: CoroutineScope,
+    private val dispatcherProvider: DispatcherProvider,
+    private val logout: Lazy<Logout>
```
hamorillo, line 18: ⚠️ I'm using a lazy injection here to break the circular dependency:

`ProfileService -> UnauthorizeInterceptor → LogoutUseCase → RealProfileRepository → ProfileService`

I've considered other options like:

- Refactoring LogoutUseCase to depend on `ProfileDao` instead of `ProfileRepository` so that it would break the circular dependency.
- Splitting the `ProfileRepository` 

But, those are more complex to implement and later on to understand when reading the code.

Using the `Lazy` injection preserves the current architecture without changes while breaking the circular dependency.

I would be happy to read opinions about this approach.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/UserComponentModule.kt
```diff
@@ -43,8 +47,14 @@ val userComponentModule = module {
     factoryOf(::UploadAvatarUseCase) { bind<UploadUserAvatar>() }
     factoryOf(::WordPressClient)
     singleOf(::InMemoryUserSessionPersistence) { bind<UserSessionPersistence>() }
-    single { ProfileService() }
-    single { AvatarService() }
+    single { UnauthorizeInterceptor(get<CoroutineScope>(), get<DispatcherProvider>(), lazy { get<Logout>() }) }
+    single {
+        OkHttpClient.Builder()
```
hamorillo: ⚠️ ❓ <del> Instantiating our OkHttpClient with the UnauthorizeInterceptor is a straightforward way to solve the problem of handling 401 responses on ProfileService and AvatarService. In fact, when we designed the Gravatar library, that's one of the reasons we kept the door open to pass a different `OkHttpClient` than the internally created one.

<del>However, using this approach results in the loss of some other nice interceptors [we add internally in the SDK](https://github.com/Automattic/Gravatar-SDK-Android/blob/ae5eadc7c95a757642d5d0a254bbdedeaeed93bb/gravatar/src/main/java/com/gravatar/di/container/GravatarSdkContainer.kt#L38). These  can't be used hereinterceptors are valid and conflict-free, but since they are internal, they. 

<del>I'm considering opening a PR to expose those interceptors and/or the `OkHttpClient.Builder`. We can always copy/paste the files, but any other third-party app can suffer the same and could take advantage of the `OkHttpClient.Builder` being public.

<del>I would like to discuss this before opening the PR for review, so I'm looking forward to opinions. cc: @AdamGrzybkowski </del>

**Edited:** What I wrote here is wrong. We do keep the interceptors from the SDK.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
AdamGrzybkowski: The way I see it is that Logout is not necessarily the same as invalidating the user session when we receive a 401 code. Although it may look like that currently, when we get the 401, we mostly want to clean the app storage, and we wouldn't perform any network requests at that point. I think it's okay to introduce a new use case and clean the necessary things there without relying on ProfileRepository, which can perform network requests. 

I'm also okay with the current code, given the time constraints we have.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/UserComponentModule.kt
```diff
@@ -43,8 +47,14 @@ val userComponentModule = module {
...
+            .addInterceptor(interceptor = get<UnauthorizeInterceptor>())
+            .build()
+    }
+    single { ProfileService(okHttpClient = get<OkHttpClient>()) }
+    single { AvatarService(okHttpClient = get<OkHttpClient>()) }
```
AdamGrzybkowski: ⛏️  Could we move all the network-related stuff to a separate module (new or `httpClientModule`)?

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
```diff
@@ -0,0 +1,34 @@
...
+) : Interceptor {
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+        val response = chain.proceed(request)
+
+        if (response.code == HttpResponseCode.UNAUTHORIZED) {
+            applicationScope.launch(dispatcherProvider.io) {
```
AdamGrzybkowski: ⛏️ Not sure we should specify the dispatcher here. Interceptor doesn't know what logout will do; in fact, logout doesn't know which dispatcher to use as well.

AdamGrzybkowski (approved):
Looks good, left some non-blocking comments.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
hamorillo: > we mostly want to clean the app storage

As I see it, part of that "clean" is to remove the profile saved in the DB. We can create a different use case which, under the hood, will perform similar actions to the Logout one. To perform the profile removal from the DB, the new use case has to rely on the `ProfileRepository` (same case that we have now) or use the `ProfileDao`, and that is exactly what I was trying to avoid. I think `ProfileDao` should be accessed through the Repository.

It's a similar approach to what I mentioned:

> Refactoring LogoutUseCase to depend on ProfileDao instead of ProfileRepository so that it would break the circular dependency.

But creating a new use case. 

If you don't have a strong disagreement, I would keep it for now.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/data/interceptors/UnauthorizeInterceptor.kt
hamorillo: Done!

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/di/UserComponentModule.kt
hamorillo: Done!

PR #56, opened 2025-07-16
Improve the avatar URL handling in the app

AdamGrzybkowski:
### Description

This PR does two things:
1. Request a specific size of the user avatar. The default is 80x80px, which is not enough for the expanded headers.
2. Changes the logic and uses `imageId` as the cache buster when saving, uploading images. This increases the number of times we use the locally cached image, even when reselecting. 

### Testing Steps
1. Verify the avatar works as expected; you can monitor the network to confirm caching is used more often.

### DIFF
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/SelectUserAvatarUseCase.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/SelectUserAvatarUseCase.kt
@@ -1,19 +1,17 @@
 package com.gravatar.app.usercomponent.domain.usecase
 
-import com.gravatar.app.clock.AppClock
 import com.gravatar.app.usercomponent.data.AvatarCacheBusterStorage
 import com.gravatar.app.usercomponent.domain.repository.UserRepository
 
 internal class SelectAvatarUseCase(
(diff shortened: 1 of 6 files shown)


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/SelectUserAvatarUseCase.kt
```diff
@@ -1,19 +1,17 @@
...
     private val userRepository: UserRepository,
     private val avatarCacheBusterStorage: AvatarCacheBusterStorage,
-    private val clock: AppClock,
 ) : SelectUserAvatar {
 
     override suspend fun invoke(avatarId: String): Result<Unit> {
         return userRepository.selectAvatar(avatarId)
             .onSuccess {
-                avatarCacheBusterStorage.saveAvatarCacheBuster(clock.now().toString())
+                avatarCacheBusterStorage.saveAvatarCacheBuster(avatarId)
```
hamorillo, line 14: Great idea! I tried to find a case that could create some issues, but I haven't been able to come up with anything. Awesome! 💯

hamorillo (approved):
Great improvement! 💯 🚀

PR #57, opened 2025-07-17
Expandable header in the Gravatar tab

AdamGrzybkowski:
### Description

This PR implements the expandable header for the Gravatar tab. The whole concept is based on Google's TopBar scrolling behavior within a Scaffold component. 

https://github.com/user-attachments/assets/5e97c97e-4dee-419a-8f0a-8afc0617bfe4

### Testing Steps
1. Launch the app
2. Play with the header


hamorillo (approved):
Great job! 👏 🚀 

Just to keep track, because I don't think it's blocking. I would say there is a tiny shake of the menu item when scrolling slowly. You need to really look at it because, in normal use, it's almost impossible to see.  :shipit:

PR #58, opened 2025-07-17
LoginScreen adjustments

AdamGrzybkowski:
### Description

This PR adjusts the LoginScreen with:
- `Try with another account` behavior - it no longer starts the OAuth flow automatically, it only reset the view state. 
- The main button has fully rounded corners

### Testing Steps
1. Launch the app, log out
2. Confirm the button shape matches the design
3. Log in, but before approving the authorization, mock the profile response to fail
4. Tap `Try with another account` button
5. Confirm the state is restored


loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginViewModel.kt
```diff
@@ -41,10 +41,10 @@ internal class LoginViewModel(
         viewModelScope.launch {
             _uiState.update {
                 it.copy(
-                    error = null
+                    error = null,
+                    isLoading = false,
                 )
             }
-            sendAction(LoginAction.StartOAuth)
```
hamorillo, line 47: ⛏️  Not directly related to this PR, but notice this as we remove the action, and the test file is not modified. We can add some tests to cover `OnTryAnotherAccountClicked` and `OnLoginClicked` events.

hamorillo (approved):
🚀

loginUi/src/main/kotlin/com/gravatar/app/loginUi/presentation/login/LoginViewModel.kt
AdamGrzybkowski: Done!

PR #60, opened 2025-07-17
Lock orientation to portrait

AdamGrzybkowski:
### Description

As the title says. 

This will be ignored on large devices with Android 16+, but it will still be respected on phones.

### Testing Steps
1. Launch the app on the phone. 
2. Rotate the device. 
3. Confirm the orientation is locked at portrait.


hamorillo (approved):
Not a big fan of this. But we'll work on removing it later 😄 

🚀

PR #61, opened 2025-07-17
Expandable header in Profile tab

hamorillo:
### Description

This PR implements the expandable header in the Profile Tab. It uses the `scrollState` to animate the different elements of the header. 

Some points to take into account:

- We are still using the static header as it contains the save button and saving state. This will be refactored in a different PR when we implement the new save header.
- The collapsed state should always be used when the keyboard is visible
- While the user has unsaved changes, we should keep the collapsed state (as we want to show the static header)
- The profile link should open an external browser 

https://github.com/user-attachments/assets/38d5d3ff-a916-4d0b-85f1-b7a860042b8b

### Testing Steps

1. Launch the app
2. Play with the header
3. Open the keyboard and verify that the header is collapsed
4. Close the keyboard and verify the header is expanded when necessary
5. Save any changes in your profile and verify that the header is expanded if needed
6. Increase the font size and verify that everything looks good


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
```diff
@@ -76,6 +85,25 @@ internal fun ProfileScreen(viewModel: ProfileViewModel = koinViewModel(), snackb
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 internal fun ProfileScreen(uiState: ProfileUiState, onEvent: (ProfileEvent) -> Unit) {
+    val isKeyboardOpen by rememberImeState()
```
AdamGrzybkowski: ❓ Is there a reason you haven't used `    val isKeyboardOpen = WindowInsets.isImeVisible`?

(For the future) We could improve it and tie the ime insets with the expanded fraction to synchronize the animation with the keyboard.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/GravatarAvatarWithShadow.kt
```diff
@@ -33,7 +33,7 @@ internal fun GravatarAvatarWithShadow(
         mutableStateOf(false)
     }
 
-    val finalModifier = remember(loaded, borderShape) {
+    val finalModifier = remember(loaded, borderShape, modifier) {
```
AdamGrzybkowski, line 36: ❓ Is that necessary?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/ProfileScreen.kt
hamorillo: > Is there a reason you haven't used     val isKeyboardOpen = WindowInsets.isImeVisible?

No. Updated as that's simpler.

> (For the future) We could improve it and tie the ime insets with the expanded fraction to synchronize the animation with the keyboard.

Agree. Let's not add more to this PR as it's big enough, but definitely a good point for improvement.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/GravatarAvatarWithShadow.kt
hamorillo: Yes. We are modifying the avatar size with the animations, so the modifier should be part of the key.

AdamGrzybkowski (approved):
:shipit:

PR #62, opened 2025-07-17
Add a banner when no avatar is selected

AdamGrzybkowski:
### Description

An extra view to notify the user that none of the current avatars are selected. 

no_avatar

### Testing Steps

1. Launch the app
2. Make sure you have at least 2 avatars
3. Select avatar nr 1
4. Delete avatar nr 1
5. Confirm the banner is visible


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/AvatarsGridSection.kt
```diff
@@ -64,18 +52,72 @@ internal fun LazyGridScope.avatarsGridSection(
...
+            }
+        }
         items(items = avatars, key = { it.avatarId }) { avatarModel ->
             SelectableAvatar(
                 avatar = avatarModel,
                 size = avatarSize,
-                modifier = Modifier.animateItem(),
+                modifier = Modifier
+                    .animateItem()
+                    .padding(top = 10.dp),
```
hamorillo: ❓ Did you add this padding on purpose? It's not in the designs.

AdamGrzybkowski: 🤦 No, this should be elsewhere. Thanks for spotting this!

AdamGrzybkowski: Fixed.

hamorillo (approved):
🚀

PR #63, opened 2025-07-18
Theme colors

AdamGrzybkowski:
### Description

I've got the generated colors from Shaun - there were some issues with them that I tried to fix manually (primary color was wrong) and since Shaun is AFK, I think it's best to go with it for now and adjust later if needed. 

I've also adjusted some the colors of the bottom nav bar item and some text colors on the GravatarScreen.

### Testing Steps

1. Launch the app and check the colors in both dark and ligth themes.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/NewAvatarButton.kt
```diff
@@ -35,8 +35,12 @@ fun NewAvatarButton(
     Surface(
         onClick = onClick,
         contentColor = MaterialTheme.colorScheme.primary,
+        color = if (isSystemInDarkTheme()) {
```
hamorillo, line 38: 🤔 I would say that the best part of defining themes is to avoid this kind of `ifs`. I would expect to use the same color at least, for example, `surface`. I know this is not an issue of the PR, just some thoughts.

hamorillo (approved):
🎨 🚀

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/NewAvatarButton.kt
AdamGrzybkowski: hmm those buttons have completely different styles in dark/light mode so I don't think we can get away from IFs here 🤔 They do have different background colors, and borders, etc.

PR #65, opened 2025-07-18
[Tooling] Add Ruby config and PR Prototype Build

iangmaia:
Fixes AINFRA-946

### Description

This PR adds the initial Ruby / Fastlane setup for the project and adds prototype builds as comments to Pull Requests.

### Testing Steps

This PR should have a @wpmobilebot comment with a QR code and a link to an `.apk`.


.buildkite/pipeline.yml
```diff
@@ -4,37 +4,38 @@
...
       - github_commit_status:
-          context: "Gradle Wrapper Validation"
+          context: Gradle Wrapper Validation
 
   # Wait for Gradle Wrapper to be validated before running any other jobs
   - wait
 
-  - label: "⚒️ Build"
+  - label: ⚒️ Build
+    if: build.pull_request.id == null
```
AliSoftware: Why the introduction of this additional condition with `== null` (i.e. only if not a PR)? Wouldn't we want to run Builds on PRs, as opposed to only build if we're on a long-running branch but not a PR?

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+      pr_number: Integer(ENV.fetch('BUILDKITE_PULL_REQUEST', nil)),
+      reuse_identifier: 'app-prototype-build-link'
+    }
+
+    comment_on_pr(
+      **comment_params,
+      body: '🚧 Prototype build will be available soon'
+    )
+
+    prototype_build_type = 'debug'
```
AliSoftware: This could probably be a constant at the top of the file (near the `PROTOTYPE_BUILD_DOMAIN = ;` one) instead of a local variable?

fastlane/Fastfile
```diff
@@ -0,0 +1,83 @@
...
+
+UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Helper.bundler?
+
+########################################################################
+# Constants
+########################################################################
+
+PLAY_STORE_TRACK_BETA = 'beta'
+PLAY_STORE_TRACK_PRODUCTION = 'production'
+
```
AliSoftware: Those seems unused. Sure they might get useful in a future iteration/evolution of the `Fastfile`, but imho we should only add them if/when we'll start adding lanes that will need it, otherwise it's YAGNI 😉

AdamGrzybkowski:
The QR works, but in order to make the App work, we need to set up mobile secrets, as we do have a few properties that have to be loaded when building the app. I tried to run `configure-setup`, but I'm not sure whether I should first set up the folder structure inside the mobile-secrets repo or finish the above command?


Edit: ~@iangmaia I tried to set it up and push the changes to this PR but sth isn't working. The `configure_apply` failed with `Decryption key could not be found`. I tested it locally, and it worked fine 🤔~

Edit2: Fixed

.buildkite/pipeline.yml
```diff
@@ -56,14 +57,24 @@ steps:
...
       ./gradlew verifyRoborazziDebug -Pscreenshot
     artifact_paths:
       - "**/build/test-results/*/*.xml"
     notify:
       - github_commit_status:
-          context: "Screenshot tests"
\ No newline at end of file
+          context: Screenshot tests
+
+  - label: 🏗️ Prototype Build
```
wzieba: I wonder: can we make this "on demand"?

I wanted to suggest it for other projects, but as we introduce prototype builds in Gravatar, I thought it's a good place to propose it.

My motivation is that a significant portion of PRs doesn't need prototype builds. They are useful when they're needed, but most of the time they waste our resources and add additional noise to a PR.

I think we should run this step on demand, like we do with Terraform builds.

AdamGrzybkowski: That sounds reasonable to me! I rarely use those builds, so if I have to ask for the build when I need it, that's fine for me.

.buildkite/pipeline.yml
iangmaia: Ah, I should have added a comment in the file or at least in GitHub -- this is the same build we use to make a prototype, so it kinda seemed redundant when there is a PR. Then this job would still run when for instance we merge a PR to `trunk` and there's no PR.

With @wzieba's [suggestion](https://github.com/Automattic/Gravatar-Android/pull/65#discussion_r2218371300) to make the Prototype PR build on demand, perhaps we could remove this condition (though it will still  feel wasteful to run it when the prototype also ran 🤷 ).
OR we could just leave it as is (always having the prototype, with the regular build running on merges on `trunk`) as it will be a way of checking the successful build.

.buildkite/pipeline.yml
ParaskP7: Great question to ask @wzieba ! 💯

> My motivation is that a significant portion of PRs doesn't need prototype builds. They are useful when they're needed, but most of the time they waste our resources and add additional noise to a PR.

A follow-up, say we make prototype build run on demand, which will avoid us building say `assembleDebug`. Wouldn't we still need to run such a job? And if we do, wouldn't that take the same amount of time, effectively defeating the purpose of making prototype builds run on demand? Or is it, that the non-assemble part of a prototype build is what takes most of the time? 🤔

PS: Now, I understand that on projects like [PCAndroid](https://github.com/Automattic/pocket-casts-android), we run both, a [Prototype Builds](https://github.com/Automattic/pocket-casts-android/blob/main/.buildkite/pipeline.yml#L90-L96) and an [Assemble release APKs](https://github.com/Automattic/pocket-casts-android/blob/main/.buildkite/pipeline.yml#L70C13-L88) job, but this is done for a different need, running assemble on release, so not sure... 🤷

iangmaia: Yep, I think we can do something like that 👍 there's a brief [discussion](https://github.com/Automattic/Gravatar-Android/pull/65#discussion_r2218426073) though on the regular Build vs. Prototype build, perhaps you folks can also chip in 🙂

iangmaia:
> The QR works, but in order to make the App work, we need to set up mobile secrets, as we do have a few properties that have to be loaded when building the app. I tried to run `configure-setup`, but I'm not sure whether I should first set up the folder structure inside the mobile-secrets repo or finish the above command?
> 
> Edit: ~@iangmaia I tried to set it up and push the changes to this PR but sth isn't working. The `configure_apply` failed with `Decryption key could not be found`. I tested it locally, and it worked fine 🤔~
> 
> Edit2: Fixed

Hey @AdamGrzybkowski, thanks for that! I was aware the `configure` step would still be needed but was leaving it to another PR... but given the changes aren't that big and they seemed to work already, I'm totally fine with it 👍

fastlane/Fastfile
iangmaia: Ah, I removed some part of the code I was already working on with these builds but didn't remove the constants. Will clean them up to add them in a future PR 👍

fastlane/Fastfile
iangmaia: Updated on 351c7f3081ceccd00325b6ae770736635ac7aee5.

fastlane/Fastfile
iangmaia: Updated on 351c7f3081ceccd00325b6ae770736635ac7aee5.

.buildkite/pipeline.yml
wzieba: > And if we do, wouldn't that take the same amount of time, effectively defeating the purpose of making prototype builds run on demand?

Maybe we do need, but it doesn't "effectively defeating the purpose of making prototype builds run on demand".

Storing prototype builds in the cases when no one uses them:
- unnecessarily increases costs of S3 hosting
- increases noise on the PR

----

I'd propose:
- running `assembleDebug` on CI checks to verify if the app compiles
- using `assembleRelease` for prototype builds on demand. It's especially good strategy IMO when we talk about the app that uses Jetpack Compose, which is significantly faster after R8.

ParaskP7: > I'd propose:

I like this proposal, I think we can try and experience with it! 👍

iangmaia: I've implemented these changes on 0e2bc539e73364c3437fb6aa93bb11f7b788e6a5:
- `assembleDebug` as a CI check (as it was before)
- `assembleRelease` for the Prototype PR build
- Prototype build only on-demand, triggered via Buildkite

This is how it looks on Buildkite:

wzieba: Looks good to me, thank you @iangmaia !

AliSoftware: If we make Prototype Builds only on-demand, I'd argue that it might be worth it to still post a PR comment to mention that they are available.

One of the reasons we made Prototype Build be available on every PR without any on-demand process was also for developers and designers to actually know those were available in the first place, and also to incentivize people to actually test PR changes at runtime (instead of only looking at the PR diff without testing).

If we do them on demand but there's nothing announcing that they are available (and only people who read the Field Guide or P2 or wherever we'd document that would know how to request the CI to generate them):
 - People will not know we have that feature of Prototype Builds anymore (especially after some time and people rotating teams or designers not being used to interact with PRs in the repo but suddenly needing to test design changes for a feature devs are working on, etc…)
 - People will be less incentivized to use them, and tend to only review the code diff without testing by installing the Prototype Build of a PR to play with the changes in the app

I agree that they are not always necessary though:
 - It might be too much to generate them on ever single commit, especially when e.g. people are working iteratively on their PR and only once they've pushed their last commit and consider the PR ready for review might it be worth it to generate a Prototype Build for example
 - Some trivial changes might not be worth a Prototype Build
 - Some devs prefer to checkout the branch and compile the code from the PR's branch in their Android Studio locally rather than using the already-precompiled-and-ready-to-install Prototype Build

So making them on-demand can still make sense… but maybe posting a PR comment like `"If you want to generate a Prototype Build, go to <link-to-buildkite-build> and unblock the «On Demand Prototype Build» action"` during a step that the On-Demand step depends on would help get the best of both worlds?

app/proguard-rules.pro
```diff
@@ -18,4 +18,16 @@
 
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
-#-renamesourcefileattribute SourceFile
\ No newline at end of file
+#-renamesourcefileattribute SourceFile
+
+# R8 missing class warnings for inter-module dependencies
```
iangmaia: @AdamGrzybkowski can you please review the proguard changes I've pushed on 4222d6aafc79c633b63bd8755ed7c8dd9a7b0b8e?

The `assembleRelease` build [initially failed](https://buildkite.com/automattic/gravatar-android/builds/280#01982c91-d900-447d-986b-8fdd18a3879c/249-1179). then I figured it was a proguard issue and tried to see if I could fix it myself with the help of Claude (it seems [it worked](https://buildkite.com/automattic/gravatar-android/builds/283#01982d56-48d6-4a38-a21d-facdb8910bc0/649)).

wzieba: it's strange, we shouldn't need to worry about internal modules obfuscation

.buildkite/pipeline.yml
wzieba: I would vote against an additional PR comment to not make more noise on PRs. I don't think it's that important for each PR to add the same information. The existence of prototype builds can be shared via some internal-team systems or in a static README or something. But ultimately, I guess it's a decision of the team.

iangmaia: Good points @AliSoftware @wzieba. I'll leave that up to the team.
If we add it, we could do it via a Danger message (and that requires the Danger setup, which I can add in another PR anyway), but I understand the argument of avoiding even more clutter.

I'll leave that up to the team (cc @AdamGrzybkowski), and we can make that choice in a follow-up PR.

app/proguard-rules.pro
AdamGrzybkowski: I'm looking into it. So far, it looks like this is caused by Koin and how it works under the hood. R8 removes a bunch of code that shouldn't be removed, and those `dontwarn` are actually not correct because the app will crash at runtime.

AdamGrzybkowski: I opened a [PR](https://github.com/Automattic/Gravatar-Android/pull/70) that should fix the release build. Once merged, you can rebase/merge the trunk.

.buildkite/pipeline.yml
AdamGrzybkowski: On-demand prototype builds work for me. I never use them for testing, I build the app from AS anyway so I only use them if I want to share the .apk with someone that can't build the app. 

How about you @hamorillo?

app/proguard-rules.pro
AdamGrzybkowski: The PR was merged.

.buildkite/pipeline.yml
hamorillo: I always build the app from AS, too, so on-demand also works for me. 

About the comment, I understand the reasons @AliSoftware and those are very valid. However, in our case (Gravatar-Android), I don't expect many people (apart from the developers) to generate a build proactively. The direction will be the opposite: developers asking to test a concrete build so we can skip the comment. We can still detail the process in the README, though.

AdamGrzybkowski:
@iangmaia The build was successful now but the apk can't be installed due to missing certificates (at least that's what AS showed me).

wzieba:
👋 I allowed myself to chime in with 1a26327.

The reason for the "missing certificate" error and similar is that we didn't have a `signingConfig` set for `release` build type. For this reason, on current `trunk` you can't even run `:app:installRelease` task - it doesn't exist.

As a temporary solution, to unblock this PR, I suggest setting the defaults debug `signingConfig` for `release`. This will make the prototype build installable on real devices.

AdamGrzybkowski (approved):

hamorillo:
The On Demand is working as expected, but it's a bit confusing because in the PR, we don't see a 🟢  as there is a "pending" check.

image

❓ Maybe it is not possible, but could we make the 🟢 not depend on the Prototype Build step?

iangmaia:
@hamorillo 
> ❓ Maybe it is not possible, but could we make the 🟢 not depend on the Prototype Build step?

Hmm I'm not 100% sure if that's possible. It's not marked as a required check, so it's not a blocker for merging, but indeed the yellow state is annoying. Not sure if there's a solution for that 🤔 I don't know it for sure but I'll have a look at it later.

AliSoftware:
@iangmaia I guess we could also consider just not reporting the `github_commit_status` of the Prototype step so it doesn't even appear in Github PR checks? (Would not help with discoverability of the on-demand Prototype Build feature though…)

PR #66, opened 2025-07-19
Keep VMs alive for the main bottom tab destination 

AdamGrzybkowski:
### Description

What:
The UI state on the profile screen is lost after using the system back button.

Why:
Using a back button on the share or the profile tab pops up the destination, and that destroys the ViewModel related to that destination. 

How:
Create a ViewModelStoreOwner for the root navHost and pass it to each destination for ViewModel creation. 

https://github.com/user-attachments/assets/d93f4fa9-3fa1-4598-bbf8-a211c65cb5f9

### Testing Steps
1. Launch the app
2. Open the Profile tab and make some changes (Don't save them)
3. Using the system back button, go back to the first tab
4. Go back to the Profile tab
5. Confirm the changes are still there and the save button is visible


hamorillo (approved):
🚀

PR #67, opened 2025-07-21
Improve lint setup: check `app` subprojects (dependencies)

wzieba:
Closes: AINFRA-969

### Description

As I was trying to enable `DoNotMockDataClass` rule, I was blocked by an issue on slack-lints projects (https://github.com/slackhq/slack-lints/issues/361). For now, it's not possible to enable this rule if we use `MockK` in this project.

But anyway, I've used the opportunity to improve the setup of lint in this project. Instead of running `lintDebug` for every module and getting multiple reports, the `app` has now `checkDependencies=true` - meaning, that `lintDebug` of this module will verify all subprojects as well and prepare the common report.

You can read more about this strategy in [the docs](https://googlesamples.github.io/android-custom-lint-rules/user-guide.md.html) under "Only analyze app/leaf modules" section

### Testing Steps

Green ✅ from CI should be enough.


AdamGrzybkowski (approved):
Thanks!

wzieba:
I see 

```JSON
{
   "message":"Server Error"
}
```

on SARIF file upload from Detekt and "Code scanning results / Detekt" is in progress for 12 minutes (it should take 2s).

Does it happen often @AdamGrzybkowski or not really? I'll force-push to restart these checks.

PR #68, opened 2025-07-21
No Internet banner on HomeScreen

AdamGrzybkowski:
### Description

In this PR:

- A new module `:networkMonitor` is created that exposed `NetworkMonitor` interface and implements `AndroidNetworkMonitor` to listen for connectivity changes within the app.
- No internet banner is shown on HomeScreen based on the `NetworkState` from the `NetworkMonitor`.

no_internet

### Testing Steps

1. Launch the app
2. Disable the network (airplane mode works)
3. Verify the banner is visible
4. Enable the network
5. Verify the banner is gone


app/src/main/java/com/gravatar/app/di/AppModule.kt
```diff
@@ -13,5 +14,6 @@ val appModule = module {
         analyticsModule,
         dispatcherModule,
         clockModule,
+        networkMonitorModule,
```
AdamGrzybkowski, line 17: We should only have one `AndroidNetworkMonitor`, so the `app` module should be responsible for adding the Koin module to the DI graph.

hamorillo: I think adding the `AndroidNetworkMonitor`, but just mentioning that Koin ensures that only one is created, despite the module being added to different places.

image

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
```diff
@@ -0,0 +1,93 @@
...
+                activeNetworks.add(network)
+                emit(NetworkState.CONNECTED)
+            }
+
+            override fun onLost(network: Network) {
+                activeNetworks.remove(network)
+                if (activeNetworks.isEmpty()) {
+                    emit(NetworkState.DISCONNECTED)
+                }
+            }
```
AdamGrzybkowski: The `onLost` callback can be invoked multiple times for each `network`. That means we can't use this callback for direct `NetworkState` conversion; we have to monitor active networks and only emit `DISCONNECTED` when the list is empty.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
```diff
@@ -0,0 +1,75 @@
...
+        val networkRequest = NetworkRequest.Builder()
+            .addCapability(NET_CAPABILITY_INTERNET)
+            .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+            .addTransportType(NetworkCapabilities.TRANSPORT_ETHERNET)
+            .addTransportType(NetworkCapabilities.TRANSPORT_CELLULAR)
+            .build()
+
+        connectivityManager?.registerNetworkCallback(networkRequest, networkCallback)
+    }
+
```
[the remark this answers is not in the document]
AdamGrzybkowski: FYI. This should be fine as the `NetworkMonitor` runs as long as the app does, but I will look into changing it to react to `subscriberCount`.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
```diff
@@ -0,0 +1,93 @@
...
+    private val state = MutableSharedFlow<NetworkState>(replay = 1)
+
+    private val activeNetworks = mutableSetOf<Network>()
+
+    init {
+        emitInitialState()
+        registerNetworkCallback()
+    }
+
+    private fun registerNetworkCallback() {
```
hamorillo, line 33: ⚠️ I think this works in the general case; however, there are some cases where we'll emit a wrong state. For example:

- WIFI is enabled and a proxy is configured, but the proxy is not working
- CELLULAR is enabled and working

If I disconnect CELLULAR we won't emit the `DISCONNECTED` state as we still have WIFI in the `activeNetworks` set. 

We should only add to `activeNetworks` the ones that:
`hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)`, receiving the `onAvailable` callback, is not enough to add them to the list.

Therefore, I believe we need to refine the logic to capture those cases.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
```diff
@@ -0,0 +1,93 @@
...
+                    emit(NetworkState.CONNECTED)
+                } else {
+                    emit(NetworkState.DISCONNECTED)
+                }
+            }
+        }
+        val networkRequest = NetworkRequest.Builder()
+            .addCapability(NET_CAPABILITY_INTERNET)
+            .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+            .addTransportType(NetworkCapabilities.TRANSPORT_CELLULAR)
```
hamorillo, line 64: ❓ Why did you remove ETHERNET?

hamorillo (commented):
The UI looks great. I think we need to improve the logic in `AndroidNetworkMonitor`.

app/src/main/java/com/gravatar/app/di/AppModule.kt
AdamGrzybkowski: I know and I don't like that because it encourages not caring about the app's DI graph and delegating it to Koin. If Koin changes the implementation (it won't, but still), then our graph will be broken.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
AdamGrzybkowski: It felt unnecessary for mobile phones. Do you think we should have it?

hamorillo: I don't expect the app to be used a lot on devices other than Phones. However, given the wide variety of Android devices, I would include it as it won't hurt.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
AdamGrzybkowski: According to the docs, we should avoid that 🤔 

>Do NOT call getNetworkCapabilities(Network) or getLinkProperties(Network) or other synchronous ConnectivityManager methods in this callback as this is prone to race conditions (there is no guarantee the objects returned by these methods will be current). Instead, wait for a call to onCapabilitiesChanged(Network, NetworkCapabilities) and onLinkPropertiesChanged(Network, LinkProperties) whose arguments are guaranteed to be well-ordered with respect to other callbacks.

AdamGrzybkowski: @hamorillo I've reworked the `AndroidNetworkMonitor` to keep the active network along with actual internet access. We have fewer properties now, so I think this should be less error-prone. The `end` `NetworkState` is based directly on the active networks that are kept as a `MutableStateFlow`. I no longer set the `NetworkState` directly from the NetworkCallback; the callbacks only modify the set of networks.

This also adheres to the suggestion from the documentation.

AdamGrzybkowski: Can you verify if this covers your scenario?

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
AdamGrzybkowski: Added it back 👍

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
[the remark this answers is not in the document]
AdamGrzybkowski: @hamorillo Here's the commit with the `subscriptionCount` https://github.com/Automattic/Gravatar-Android/pull/68/commits/60422bedc8a8efd13e779f87a716f1dec8db0cda

Personally, I would be ok without this - given that we have an app wide monitor, the cleanup will be done when the app is killed but if we want to have that extra cleanup when for example the users logs out and there are no subscribes there then that would do it.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
hamorillo: > Can you verify if this covers your scenario?

It does solve the issue.

> According to the docs, we should avoid that 🤔

What I was suggesting is to rely only on the `onCapabilitiesChanged` callback, not calling any methods in `onAvailable`. In fact, I do think we don't need to implement anything on that method as it doesn't provide useful information for us. (Know that a concrete network is available; it's not enough in our case. We need to know that it's connected to the Internet).

I think my suggestion can simplify the code a bit by only keeping track of the networks that are connected to the Internet. We can do that on the `onCapabilitiesChanged` callback. We would add the network to the `set` only if it has Internet. 

I've pushed my suggestion here as an example:

https://github.com/Automattic/Gravatar-Android/compare/adam/GRA-467...hamorillo/GRA-467-suggestion

Let me know what you think.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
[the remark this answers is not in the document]
hamorillo: > Personally, I would be ok without this

I agree. In fact, it would be nice to inform the user also in the login screen that they can't perform the login because of the connectivity. (As a future improvement)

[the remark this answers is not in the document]
AdamGrzybkowski: Do you want me to remove the last commit?

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
AdamGrzybkowski: Looks good, do you want to push this commit to this branch?

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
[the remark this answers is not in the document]
hamorillo: I'm tempted to say yes because it doesn't add much value. You explained it really well [here](https://github.com/Automattic/Gravatar-Android/pull/68#discussion_r2222573389): 

> given that we have an app wide monitor, the cleanup will be done when the app is killed

However, I don't have a strong opinion; feel free to do whatever you prefer.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
hamorillo: Sure! Let me know if you plan to revert the last commit or not (as it will be easier to revert it before pushing mine). :)

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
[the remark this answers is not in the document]
AdamGrzybkowski: Reverted.

networkMonitor/src/main/kotlin/com/gravatar/app/networkmonitor/AndroidNetworkMonitor.kt
AdamGrzybkowski: After I reverted the last one, this was a clean cherry-pick so I applied it here.

hamorillo: Nice!

hamorillo (approved):
🚀

PR #69, opened 2025-07-22
TopBarMenu in GravatarScreen

hamorillo:
### Description

This PR introduces the main menu with the following actions:

- Visit your profile
- Share
- Gravatar.com
- Sign out

The menu button has been added to the Gravatar and Profile tabs.

| Gravatar | Profile |
|----------|-----------|
|image|image|

**What's not included in this PR** (it will be in separate PRs)

- About this app
<del>- Integrate menu in ProfileHeader</del>

### Testing Steps

- Launch the app
- Verify each of the actions included in the Menu (`Visit your profile`, `Share`, `Gravatar.com`, `Sign out`)
- Verify it's correctly positioned with the header expanded and collapsed
- Switch between dark/light themes

Perform the same checks in the ProfileScreen


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/TopBarPickerPopup.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.compose.ui.unit.dp
+import com.gravatar.app.homeUi.R
+import com.gravatar.ui.GravatarTheme
+
+@Composable
+internal fun TopBarPickerPopup(
+    anchorAlignment: Alignment.Horizontal,
+    offset: DpOffset = DpOffset(0.dp, 0.dp),
+    onDismissRequest: () -> Unit,
+    onTopBarOptionClicked: (TopBarOption) -> Unit,
```
AdamGrzybkowski: 💡 We will have to implement this callback on 3 different screens. Would it make sense to keep as much of the logic encapsulated in this component? I would even consider having a separate ViewModel for this composable that could handle all the events.

AdamGrzybkowski (commented):
Works well! I would consider a different approach that would be more plug-and-play since we will be using this component on multiple screens. Let me know what you think.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/TopBarPickerPopup.kt
hamorillo: Yes, I was considering that as the next step: integrating the menu into the ProfileScreen. But I can iterate over it in this PR.

AdamGrzybkowski: I think it makes sense because you're adding a bunch of code to the `GravatarViewModel` in this PR that will be removed in the next one.

hamorillo: I've pushed the changes here. Also, added the menu to the `ProfileScreen` so we can verify that it is easy to reuse.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/topbar/TopBarPickerPopupViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+    init {
+        collectUserProfile()
+    }
+
+    fun onEvent(event: TopBarPickerPopupEvent) {
+        when (event) {
+            TopBarPickerPopupEvent.OnLogoutSelected -> logoutUser()
+            TopBarPickerPopupEvent.OnProfileLinkClicked -> openProfileUrl()
+            TopBarPickerPopupEvent.OnGravatarLinkClicked -> openGravatarWebsite()
+            TopBarPickerPopupEvent.OnShareProfileClicked -> shareProfileUrl()
```
AdamGrzybkowski, line 33: 💡 We can avoid this back and forth between view <-> viewModel if we set the `profile` in the VM UI state (or just the URLs that we are interested in) and build the click listener based on those state values.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/topbar/TopBarPickerPopup.kt
```diff
@@ -0,0 +1,159 @@
...
+        popupMenu = PickerPopupMenu(
+            items = listOf(
+                PickerPopupItem(
+                    text = stringResource(R.string.gravatar_tab_topbar_menu_visit_profile),
+                    iconRes = R.drawable.top_bar_menu_profile,
+                    contentDescription = stringResource(
+                        R.string.gravatar_tab_topbar_menu_visit_profile
+                    ),
+                    onClick = {
+                        onEvent(TopBarPickerPopupEvent.OnProfileLinkClicked)
```
AdamGrzybkowski, line 81: Here, instead of `onEvent`, we could already have the `profileUrl` passed in the uiState and do exactly what we do when the `TopBarPickerPopupAction.OpenExternalUrl` is collected.

AdamGrzybkowski (commented):
Thanks for the changes! It is a plug-and-play now! 

I left one suggestion that could remove a bunch of code and simplify it a bit.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/topbar/TopBarPickerPopupViewModel.kt
hamorillo: While it's true that we'll save a few lines of code, I still prefer the current approach. 

I know that, at the moment, the actions are pretty simple. However, the current implementation is what I would expect, and it's more testable, for example, when we introduce some analytics. In that case, we'll add the analytics tracker to the viewmodel and we'll be able to verify with a unit test that we are truly saving the analytics. If we remove the events and the actions, we'll need to inject the tracker on the UI and won't be able to easily test that it's correctly used.

AdamGrzybkowski (approved):
:shipit:

PR #70, opened 2025-07-22
Fix release build type

AdamGrzybkowski:
### Description

In this [PR](https://github.com/Automattic/Gravatar-Android/pull/65), with the prototype build, it was discovered that the R8 fails when building the release variant. 

When trying to fix it, I noticed multiple errors, conflicting names (after obfuscation) - it was a bit weird and something I haven't seen before. 

It turned out, the cause of this was the `isMinifyEnabled=true` set for each library module. As far as I know, this is best to only set this in the `app` module, as the minification will be enabled on all dependencies regardless of each individual module setup. 
 
### Testing Steps

You can test this by building the `release` variant of the app. To do so, you have to add this line:
```kotlin
signingConfig = signingConfigs.getByName("debug")
```
in [this block](https://github.com/Automattic/Gravatar-Android/blob/trunk/build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt#L43).

After that, build the app and smoke test it.

### DIFF
--- a/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
+++ b/build-logic/convention/src/main/kotlin/GravatarAndroidLibraryConventionPlugin.kt
@@ -32,7 +32,7 @@ class GravatarAndroidLibraryConventionPlugin : Plugin<Project> {
     private fun LibraryExtension.configureBuildTypes() {
         buildTypes {
             getByName("release") {
-                isMinifyEnabled = true
+                isMinifyEnabled = false
                 proguardFiles(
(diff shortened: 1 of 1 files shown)


wzieba:
> It turned out, the cause of this was the `isMinifyEnabled=true` set for each library module. As far as I know, this is best to only set this in the `app` module, as the minification will be enabled on all dependencies regardless of each individual module setup.

True 👍 we do this in other projects at A8c.

The behavior you described was introduced in 8.4.0 https://developer.android.com/build/releases/past-releases/agp-8-4-0-release-notes#library-classes-shrunk

hamorillo (approved):
🚀

PR #73, opened 2025-07-23
Share tab header with the menu button

AdamGrzybkowski:
### Description

This PR implements the layout for the Share tab header with the menu button.

Additionally, I've extracted the blurred avatar component so that we can reuse it across all the headers and added a semi-transparent black layer over it to match iOS.

This PR doesn't:
- Implement the QR code. There's a slot for it, though.
- Add all the buttons. Only menu was added for now, as it's not clear what will be implemented for the first push and we can't have buttons that don't work.

share_header

### Testing Steps

1. Open the app and go to the share tab
2. Confirm the layout looks like in Figma
3. Test other headers because of the small structure change


AdamGrzybkowski:
The screenshot tests pass locally...

hamorillo:
> The screenshot tests pass locally...

I wonder why it doesn't work in CI. In both executions, the difference is just 12 pixels in the ProfileScreenTest 🤔

hamorillo (approved):
🚀

AdamGrzybkowski:
>I wonder why it doesn't work in CI. In both executions, the difference is just 12 pixels in the ProfileScreenTest 🤔

I have updated the Roborazzi rule with the same compare options we use in the SDK - it now works fine.

PR #75, opened 2025-07-23
AboutAppDialog from the TopMenuPopup

AdamGrzybkowski:
### Description

This PR implements the `AboutAppDialog` on all tabs where the menu button is visible. 

The `AppVersion` is injected from the `:app` module.

Screenshot_20250723_212438

### Testing Steps
1. Launch the app
2. Open the menu and tap "About this app" on each tab
3. Confirm you see the dialog
4. Tap on the URLs, email to verify the click listeners
5. Dismiss it
6. Confirm it's gone


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/topbar/components/AboutAppDialog.kt
```diff
@@ -0,0 +1,194 @@
...
+import androidx.core.net.toUri
+import com.gravatar.app.design.theme.GravatarAppTheme
+import com.gravatar.app.homeUi.AppVersion
+import com.gravatar.app.homeUi.R
+import org.koin.compose.koinInject
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+internal fun AboutAppDialog(
+    visible: Boolean,
```
hamorillo: I don't usually like this approach, as we are adding an unnecessary composable that will always be there. Not a big deal, I understand that it's convenient in some cases.

hamorillo (approved):
Looks great! 🚀

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/topbar/components/AboutAppDialog.kt
AdamGrzybkowski: No strong opinions here, I removed it.

PR #76, opened 2025-07-23
Generate QR code with vCard data

hamorillo:
### Description

This PR implements the QR code generation using [qrose](https://github.com/alexzhirkevich/qrose). 

<details>
<summary> Before Public Fields Section </summary>

[Your text here...](https://github.com/user-attachments/assets/2ae411ae-cac2-48d9-b59c-78f50811ed19)
</details>

https://github.com/user-attachments/assets/8d155799-5478-43c9-9b2e-5a9e2c87dbbd

**What's included?**

This PR generates a QR code containing the vCard information using the Profile data, plus the private contact information introduced by the user. It also takes into account the switches in the `Private Contact Section`, so if the user decides not to share their email and/or phone, those won't be included.

**What's not included?**

<del>The public fields section is not included, but the data is being included in the QR. We need to implement the UI to show that section and take into account the switches while generating the QE</del>

_Edited:_ The public fields are now included in this PR, as they were already added. You can include those fields in the vCard by playing with the switches.

**Why I've chosen `qrose`?**
- It's just `kotlin` and `compose.ui`, it doesn't have more dependencies.
- It's easy to customize, for example:
image

### Testing Steps

- Launch the app and navigate to the Share section
- Verify the QR code is, in fact, a vCard with the profile information
- Play with the phone and email fields/switches, verifying the QR code generation is correct
- Play with the public fields/switches, verifying the QR code generation is correct


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/ShareUiState.kt
```diff
@@ -5,12 +5,62 @@ import com.gravatar.restapi.models.Profile
...
 internal data class PrivateContactInfo(
     val emailValue: String = "",
-    val isEmailShared: Boolean = false,
+    val isEmailShared: Boolean = true,
     val phoneValue: String = "",
-    val isPhoneShared: Boolean = false
+    val isPhoneShared: Boolean = true,
 )
+
+private fun generateVCardData(profile: Profile?, privateContactInfo: PrivateContactInfo): String {
```
AdamGrzybkowski: Can we create a VCard builder and abstract away the technical details so it would look more like this:

```kotlin
VCard.Builder()
    .setNickname()
    ...
```
    
This should make it easier to test the VCard generation in isolation.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/ShareUiState.kt
```diff
@@ -5,12 +5,62 @@ import com.gravatar.restapi.models.Profile
...
+    val vCardBuilder = StringBuilder()
+        .append("BEGIN:VCARD\n")
+        .append("VERSION:3.0\n")
+        .append("PRODID:Gravatar Android\n")
+
+    // Add name information if available
+    if (profile != null) {
```
AdamGrzybkowski: ❓ Have you looked into adding the avatar?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/ShareUiState.kt
```diff
@@ -5,12 +5,62 @@ import com.gravatar.restapi.models.Profile
 internal data class ShareUiState(
     val profile: Profile? = null,
     val avatarUrl: String? = null,
-    val privateContactInfo: PrivateContactInfo = PrivateContactInfo()
-)
+    val privateContactInfo: PrivateContactInfo = PrivateContactInfo(),
+) {
+    val vCardQrCodeData: String = generateVCardData(profile, privateContactInfo)
```
AdamGrzybkowski: ⛏️ I think it would be great to delay the QR generation a bit, rather than generating it on the fly every time. Just leaving it here, this doesn't have to be done in this PR.

AdamGrzybkowski (commented):
Works great! I would prefer to have the VCard generation extracted so that it can be tested in isolation, not as part of the uiState. WDYT?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/ShareUiState.kt
```diff
@@ -5,12 +5,62 @@ import com.gravatar.restapi.models.Profile
...
+}
 
 internal data class PrivateContactInfo(
     val emailValue: String = "",
-    val isEmailShared: Boolean = false,
+    val isEmailShared: Boolean = true,
```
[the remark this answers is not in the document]
AdamGrzybkowski: Is that on purpose? It's ON even with the field empty 🤔

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/components/ShareHeader.kt
```diff
@@ -33,13 +34,16 @@ import com.gravatar.app.homeUi.R
...
+import io.github.alexzhirkevich.qrose.rememberQrCodePainter
 
 @Composable
 internal fun ShareHeader(
     avatarUrl: String,
+    vCardQrCodeData: String,
     modifier: Modifier = Modifier,
 ) {
     var topBarMenuVisible by remember { mutableStateOf(false) }
+    val qrcodePainter: Painter = rememberQrCodePainter(vCardQrCodeData)
```
AdamGrzybkowski: ❓ Can we round the corners to match the design better?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/ShareUiState.kt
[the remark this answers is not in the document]
hamorillo: Yes, I think that's what we agreed in the last call, and it's how the iOS app is working at the moment. However, we can improve this behaviour. We can discuss this with iOS folks.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/ShareUiState.kt
hamorillo: I took a look at the iOS-generated file, but I haven't started with the implementation here.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/components/ShareHeader.kt
hamorillo: Done!

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/ShareUiState.kt
hamorillo: Done!

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/model/VCard.kt
```diff
@@ -0,0 +1,74 @@
...
+        val firstName = firstName.orEmpty()
+        val lastName = lastName.orEmpty()
+        val nickname = nickname.orEmpty()
+        if (firstName.isNotEmpty() || lastName.isNotEmpty()) {
+            contentBuilder.append("N:$lastName;$firstName;;;\n")
+                .append("FN:${("$firstName $lastName".trim()).ifEmpty { nickname }}\n")
+        }
+        val calculatedNickname = nickname.ifEmpty { "$firstName $lastName".trim() }
+
+        calculatedNickname.takeIf { it.isNotEmpty() }?.let { contentBuilder.append("NICKNAME:$it\n") }
```
AdamGrzybkowski: This doesn't match the iOS:

it should be:

```
  N:\(model.lastName);\(model.firstName);
  FN:\(model.displayName)
  NICKNAME:\(model.displayName)
```

hamorillo: 🤔 I'm not sure if that is the correct approach. The specification says for the FN:

> Purpose:  To specify the formatted text corresponding to the name of the object the vCard represents.

I can update the nickname so that it does not use the first and last name when it's empty. But the FN looks better as it is.

AdamGrzybkowski: @etoledom thoughts on this?

etoledom: This is the logic we used with @pinarol for this one:

`FN`: Is a required field on the vcard spec, so we chose a profile property which we know will always have data to fill it  with (`displayName`).
`N`: Will be filled with the first and last name if they exist.
`NICKNAME`: Is technically how the user wants to be called

We noticed by testing (on iOS contacts) that whenever `N` is present, `FN` gets replaced by `N`.
If N is empty and ORG is present, FN is replaced by ORG
If N and ORG are empty, FN becomes N (first and last name are split by spaces).

While `NICKNAME` is always preserved.

So, if N and ORG are empty, we will have FN = N = displayName
If N or ORG have data, FN gets replaced by either, so whatever data in FN is irrelevant.
And NICKNAME will always have the data of display name, as it is how the user wants to be recognized (or so we assume), and also, is a way to not get it replaced by N or ORG.

If FN = N (first - last name), FN becomes irrelevant in any case.

To know if this also counts for Android, we will need to know if FN is added to a separated field than N when both have different data. And a similar test when N is empty and ORG has data. Is FN replaced by ORG or is it added to a separated contact field?

hamorillo: Great explanation @etoledom! Appreciate it.

I've done some tests and I'm not wrong, what you've detailed also applies to Android. I observe the same behaviour so, I'll update the code to save `displayName` in `FN`.

Thanks!

hamorillo: Updated!

AdamGrzybkowski: I can't confirm this behavior on my device.

Now that the `displayName` is added as `FN` and `NICKNAME`, it's always added as a first name in my Google contacts. Even if `N` has data. The `N` is ignored.

When `N` is empty and `ORG` has data, the `ORG` is in the Organization field, and `displayName` is set as first name.

Here's an example showing the N is ignored.
| Data from the profile screen | Google contacts |
|---|---|
| Screenshot_20250724_164636 | signal-2025-07-24-164857 |

hamorillo: You are right @AdamGrzybkowski. I messed up things with the code while testing. If I'm not wrong (again), if FN is present, Contact apps will use that value instead of N or NICKNAME  (I've downloaded a couple of contact apps and the behaviour seems to be the same). 

The priority is first FN, and if FN is not available, then N.  I don't see Nickname being used in any case. 🤔 

If that's correct, I would revert the last commit and set FN with the first and last name as we were doing when the discussion started. If those values are not available, I would use the display name. With this approach, I believe we should achieve acceptable results regardless of the platform scanning the QR. Wdyt?

etoledom: We can adapt iOS also to play good with Android contact.

> The priority is first FN, and if FN is not available, then N. I don't see Nickname being used in any case. 🤔

So it would be:
`FN: fullName ? displayName` ?

What to do when there's no fullName and ORG is set?
> When N is empty and ORG has data, the ORG is in the Organization field, and displayName is set as first name.

Is this ok?

cc @pinarol

pinarol: It seems that `FN` is parsed incorrectly on some Android devices, because the `space` character is treated as a delimiter(which, by the way, doesn’t comply with the spec). This becomes especially problematic when names include multiple first or last names, or prefixes.

Interestingly, when we pass an empty string to `FN`, the `N` field gets used and is parsed correctly.

That puts us in a tricky spot: leaving a “required” field empty seems to fix the issue, but it also looks like we may not have much of a choice. 

We don't need to do any changes on the ORG field in the vCard template I think.

So it becomes:

```
FN:
N: First/Last name > Display name (prioritize First/Last name over Display name)
NICKNAME: Display name
ORG: SomeCompany
...
...
```

hamorillo:
@etoledom I've removed the `FN` and updated the logic of the vCard generation. Could you give it a try, and if everything looks good, a 👍 ? 🙏

pinarol:
Hey, just to give a bit of context, we didn’t completely remove the FN field, we just left it empty since it’s required. I’m not totally sure this is the best way to handle it, but I do feel like keeping the vCard format consistent across platforms is the correct approach.

hamorillo:
When the standard says `The property MUST be present in the vCard object.` I wouldn't say an empty value in `FN` is accepted. I think neither option is meeting the standard.

I'm not sure what the best approach is, so I've added the empty `FN` to maintain consistency with iOS.

etoledom (approved):
Working great as expected 🎉 

Tested reading the generated QR code in both iOS and Xiaomi

PR #83, opened 2025-07-25
[Tooling] Add app version setting using `version.properties`

iangmaia:
Fixes AINFRA-1020

### Description

This PR moves app versioning control to the `version.properties` file so that it can be more easily updated by our automation.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -17,8 +17,8 @@ val versionProperties = loadPropertiesFromFile(file("version.properties"))
 
 project.apply {
     extra.apply {
-        set("versionName", versionProperties.getProperty("versionName", "0.1"))
-        set("versionCode", versionProperties.getProperty("versionCode", "1").toInt())
+        set("versionName", versionProperties.getProperty("versionName", null))
+        set("versionCode", versionProperties.getProperty("versionCode", null).toInt())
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -10,3 +12,24 @@ plugins {
     alias(libs.plugins.ksp) apply false
     alias(libs.plugins.room) apply false
 }
+
+val versionProperties = loadPropertiesFromFile(file("version.properties"))
+
+project.apply {
+    extra.apply {
+        set("versionName", versionProperties.getProperty("versionName", "0.1"))
+        set("versionCode", versionProperties.getProperty("versionCode", "1").toInt())
```
hamorillo: ❓ Should we not hardcode those values?
```suggestion
        set("versionName", versionProperties.getProperty("versionName", null))
        set("versionCode", versionProperties.getProperty("versionCode", null).toInt())
```

hamorillo (approved):
Thanks for working on this! I left a minor comment/question; apart from that, it looks good! 🚀

build.gradle.kts
iangmaia: Agreed, no point of having a default -- `version.properties` should always be there and properly configured, otherwise things should fail.

PR #84, opened 2025-07-25
[Tooling] Add Debug / Release signing

iangmaia:
Fixes AINFRA-1021

### Description

This PR adds Release and Debug build signing configurations.

This PR is stacked on https://github.com/Automattic/Gravatar-Android/pull/83

### Testing Steps

1. Checkout this branch
2. Update your `mobile-secrets`
3. Run `bundle exec fastlane run configure_apply`
4. Check that files relative to the project were created in `~/.configure/gravatar-android`
5. Run `./gradlew assembleDebug`
6. Run `./gradlew assembleRelease`


.gitignore
```diff
@@ -10,8 +10,12 @@ secrets.properties
 *.iml
 
 # generated files
+app/release/
 build/
-/captures
+build-logic/convention/bin
+captures/
+clock/bin/
+foundations/bin/
```
hamorillo: ❓ Do we need to ignore those folders? I don't see any files added to them.

hamorillo (approved):
🚀

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
```diff
@@ -39,15 +39,47 @@ class GravatarAndroidApplicationConventionPlugin : Plugin<Project> {
...
+            project.logger.warn("Release signing configuration skipped: no secrets properties found")
+        }
+
+        signingConfigs {
+            maybeCreate("debug").apply {
+                storeFile = project.rootProject.file("debug.keystore")
+                storePassword = "android"
+                keyAlias = "androiddebugkey"
+                keyPassword = "android"
+            }
```
wzieba: Is it necessary to share a debug keystore? If not, then I'd vote to not do this - less code to maintain.

I know that in other projects we do this, but I don't see a valid reason to justify it.

.gitignore
iangmaia: `app/release/` is where build files will be located if you generated them locally.

In my setup, when I compiled the project I got some generated files in `build-logic/convention/bin` / `foundations/bin/` / `clock/bin/` 🤔 odd that you don't see them, not sure why. Maybe something particular to my setup?
Still, I'd say it's safe to ignore `**/bin/` folders?

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
iangmaia: Thanks, that was a point I wasn't fully confident about; I've seen this setup in a few other projects and thought it was the preferred way for some reason, but I'm also not sure why it would be needed. Maybe if the builds are uploaded somewhere like Firebase? Still, those should probably be `release` builds anyway.

.gitignore
wzieba: what's in `bin`? It's not generated on my machine, usually with Gradle everything like this should be in `build` directory (which is already excluded)

iangmaia: If I revert the .gitignore changes, this is what I see; I assumed those are just some post-processed generated files thus safe to ignore, but again, odd that others don't see them:
Screenshot 2025-07-28 at 14 29 13

iangmaia: Ah! I've deleted all those files and rebuild the project; it seems they don't show anymore.
Not sure why, but maybe they were being generated in the earlier project days? Not even `app/release/app-release.aab` is there, everything is inside `build/`.

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
iangmaia: Updated on ec4d36f0046379e6a04dd5aa340bf9bf2c070de3.

.gitignore
iangmaia: Updated on 1c649b678bbf0fb7981f7c5ce791da0ab9316873 as the changes don't seem necessary.

Not sure exactly what happened: I upgraded my Android Studio since I started adding some changes to Gravatar, and checked out the project in different moments. Right now, I confirmed that if I delete the files that I had on `**/bin/` and rebuild the project (I tried `assembleDebug`, `assembleRelease`, `bundleRelease`, etc), I don't see these files and all generated ones go to `build/` 👍

hamorillo: > Not sure why, but maybe they were being generated in the earlier project days? Not even app/release/app-release.aab is there, everything is inside build/.

I don't remember having those files or the `app/release` folder. That's why I was surprised to add them to `.gitignore`.

Thanks for fixing it!

❓ What's the `captures` folder used for?

iangmaia: > ❓ What's the captures folder used for?

I imagine that's related to the screenshot testing; this is how the `.gitignore` was on `trunk`, I just reverted my changes.

hamorillo: Oh, you are right. I was there before. I'll check because I think we are not using that.

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
wzieba: thanks!

PR #86, opened 2025-07-26
[Tooling] Add initial release lane setup

iangmaia:
Fixes AINFRA-947

### Description

This PR contains the initial work for building and uploading releases to the Google Play Store on CI.

I've started adding the basics of version management utilities and the upload to a GCP testing track so we can validate the entire flow and the security setup, with a lane that can be run locally and trigger the release process in CI.

In the next iterations, we can further automate release uploading to other tracks (including production), updated metadata such as the app changelog, create GitHub releases, etc, while we also mature the release lifecycle according to the project's needs.

### Testing Steps

1. Make sure you have `rbenv` setup and environment variables such as `GITHUB_TOKEN` and `BUILDKITE_TOKEN`.
2. Checkout this branch
3. Run `bundle install`
4. Run `bundle exec fastlane new_alpha_release`. This will print what the lane will do and ask for confirmation.
5. After the confirmation, the lane should commit a version bump and trigger a build on CI. It should print the URL of the build so you can follow it.
6. The version should have been uploaded to Google Play Console. It should be under `Test And Release` - `Testing` - `Internal Testing` as a new `draft` (it seems it's not possible to upload non-drafts to a draft app).

### DIFF
--- a/.buildkite/release-pipelines/new-alpha-release.yml
+++ b/.buildkite/release-pipelines/new-alpha-release.yml
@@ -1,0 +1,18 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+steps:
+  - label: ":rocket: New Alpha Release"
+    key: new-alpha-release
+    plugins: [$CI_TOOLKIT]
+    command: |
+      echo "--- :robot_face: Use bot for Git operations"
+      source use-bot-for-git
+
+      echo "--- :ruby: Setup Ruby Tools"
+      install_gems
+
+      echo "--- :rocket: Create New Alpha"
+      bundle exec fastlane new_alpha_release skip_confirm:true
+    agents:
+      queue: tumblr-metal
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -54,7 +54,7 @@ platform :android do
       release_status: release_status,
       rollout: rollout,
       skip_upload_metadata: (track != 'production'),
-      skip_upload_changelogs: true,
+      skip_upload_changelogs: (track != 'production'),
       skip_upload_images: true,
       skip_upload_screenshots: true,
       json_key_data: ENV.fetch('PLAY_STORE_RAW_JSON_KEY', nil)
@@ -111,7 +111,7 @@ platform :android do
   end
 
   lane :build_and_upload_to_play_store do |track: 'internal', release_status: 'draft'|
-    # TODO: using `internal`` track with `draft` releases by default while the release itself is in draft mode
+    # TODO: using `internal` track with `draft` releases by default while the release itself is in draft mode
     build_bundle
     upload_to_store(track: track, release_status: release_status, rollout: release_status == 'draft' ? nil : '1')
   end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -19,9 +19,9 @@ VERSION_FILE = Fastlane::Wpmreleasetoolkit::Versioning::AndroidVersionFile.new(v
 
 VERSION_CALCULATOR = Fastlane::Wpmreleasetoolkit::Versioning::SemanticVersionCalculator.new
 VERSION_FORMATTER = Fastlane::Wpmreleasetoolkit::Versioning::RCNotationVersionFormatter.new
-# Using DerivedBuildCodeFormatter to derive build codes from version name (format: 1XXYYZZNNN)
+# Using DerivedBuildCodeFormatter to derive build codes from version name (format: 1XXYYYZNN)
 # This eliminates the need for manual build code incrementing and provides deterministic versioning
-BUILD_CODE_FORMATTER = Fastlane::Wpmreleasetoolkit::Versioning::DerivedBuildCodeFormatter.new
+BUILD_CODE_FORMATTER = Fastlane::Wpmreleasetoolkit::Versioning::DerivedBuildCodeFormatter.new(prefix: '1', major_digits: 2, minor_digits: 3, patch_digits: 1, build_digits: 2)
 
 PROTOTYPE_BUILD_DOMAIN = 'https://cdn.a8c-ci.services'
 RELEASE_BUILD_TYPE = 'Release'
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -122,32 +122,33 @@ platform :android do
     require_env_vars!('BUILDKITE_TOKEN')
     ensure_git_status_clean
 
-    Fastlane::Helper::GitHelper.checkout_and_pull(DEFAULT_BRANCH)
-
     UI.important <<~PROMPT
       Initiating a new alpha release.
       This will:
+      - Checkout and pull the #{DEFAULT_BRANCH} branch (if running in CI)
       - Bump the build code from #{build_code_current} to #{build_code_next}
       - Trigger a new alpha build in Buildkite
     PROMPT
     next unless skip_confirm || UI.confirm('Continue?')
 
+    Fastlane::Helper::GitHelper.checkout_and_pull(DEFAULT_BRANCH) if is_ci?
+
     bump_build_code
 
     push_to_git_remote
 
-    trigger_alpha_build
+    trigger_release_build
   end
 
   # Trigger a new alpha build on CI
   #
-  lane :trigger_alpha_build do
+  lane :trigger_release_build do
     require_env_vars!('BUILDKITE_TOKEN')
 
     trigger_buildkite_build(
-      branch: DEFAULT_BRANCH,
+      branch: git_branch,
       pipeline: 'release-builds.yml',
-      message: "Alpha Build (#{build_code_current})"
+      message: "Release Build (#{build_code_current})"
     )
   end
 
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -7,6 +7,11 @@ UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Hel
 ########################################################################
 # Constants
 ########################################################################
+DEFAULT_BRANCH = 'trunk'
+
+BUILDKITE_ORG_NAME = 'automattic'
+BUILDKITE_PIPELINE = 'gravatar-android'
+
 PROJECT_ROOT_FOLDER = File.dirname(File.expand_path(__dir__))
 
 VERSION_PROPERTIES_PATH = File.join(PROJECT_ROOT_FOLDER, 'version.properties')
@@ -18,7 +23,7 @@ BUILD_CODE_CALCULATOR = Fastlane::Wpmreleasetoolkit::Versioning::SimpleBuildCode
 BUILD_CODE_FORMATTER = Fastlane::Wpmreleasetoolkit::Versioning::SimpleBuildCodeFormatter.new
 
 PROTOTYPE_BUILD_DOMAIN = 'https://cdn.a8c-ci.services'
-PROTOTYPE_BUILD_TYPE = 'release'
+RELEASE_BUILD_TYPE = 'Release'
 
 platform :android do
   # Builds a release app bundle .aab
@@ -28,7 +33,7 @@ platform :android do
 
     gradle(
       task: 'bundle',
-      build_type: 'Release'
+      build_type: RELEASE_BUILD_TYPE
     )
   end
 
@@ -52,7 +57,7 @@ platform :android do
       skip_upload_changelogs: true,
       skip_upload_images: true,
       skip_upload_screenshots: true,
-      json_key_data: get_required_env!('PLAY_STORE_RAW_JSON_KEY')
+      json_key_data: ENV.fetch('PLAY_STORE_RAW_JSON_KEY', nil)
     )
   end
 
@@ -78,7 +83,7 @@ platform :android do
 
     gradle(
       task: 'assemble',
-      build_type: PROTOTYPE_BUILD_TYPE
+      build_type: RELEASE_BUILD_TYPE
     )
 
     upload_path = upload_to_s3(
@@ -94,7 +99,7 @@ platform :android do
       app_display_name: 'Gravatar Android',
       download_url: install_url,
       metadata: {
-        'Build Type': PROTOTYPE_BUILD_TYPE
+        'Build Type': RELEASE_BUILD_TYPE
       },
       fold: true
     )
@@ -105,24 +110,73 @@ platform :android do
     )
   end
 
-  def generate_prototype_build_number
-    if ENV['BUILDKITE']
-      commit = ENV.fetch('BUILDKITE_COMMIT', nil)[0, 7]
-      branch = ENV['BUILDKITE_BRANCH'].parameterize
-      pr_num = ENV.fetch('BUILDKITE_PULL_REQUEST', nil)
+  lane :build_and_upload_to_play_store do |track: 'internal'|
+    build_bundle
+    upload_to_store(track: track)
+  end
 
-      pr_num == 'false' ? "#{branch}-#{commit}" : "pr#{pr_num}-#{commit}"
+  #####################################################################################
+  # Release lanes & utils
+  #####################################################################################
+  lane :new_alpha_release do |skip_confirm: false|
+    require_env_vars!('BUILDKITE_TOKEN')
+    ensure_git_status_clean
+
+    Fastlane::Helper::GitHelper.checkout_and_pull(DEFAULT_BRANCH)
+
+    UI.important <<~PROMPT
+      Initiating a new alpha release.
+      This will:
+      - Bump the build code from #{build_code_current} to #{build_code_next}
+      - Trigger a new alpha build in Buildkite
+    PROMPT
+    next unless skip_confirm || UI.confirm('Continue?')
+
+    bump_build_code
+
+    push_to_git_remote
+
+    trigger_alpha_build
+  end
+
+  # Trigger a new alpha build on CI
+  #
+  lane :trigger_alpha_build do
+    require_env_vars!('BUILDKITE_TOKEN')
+
+    trigger_buildkite_build(
+      branch: DEFAULT_BRANCH,
+      pipeline: 'release-builds.yml',
+      message: "Alpha Build (#{build_code_current})"
+    )
+  end
+
+  def trigger_buildkite_build(branch:, pipeline:, message: nil)
+    environment = {
+      'RELEASE_VERSION' => release_version_current
+    }
+
+    common_args = {
+      pipeline_file: pipeline,
+      branch: branch,
+      message: message || "Build #{pipeline.sub('.yml', '')}",
+      environment: environment
+    }
+
+    if is_ci?
+      buildkite_add_trigger_step(**common_args)
     else
-      repo = Git.open(PROJECT_ROOT_FOLDER)
-      commit = repo.current_branch.parameterize
-      branch = repo.revparse('HEAD')[0, 7]
-
-      "#{branch}-#{commit}"
+      # For local development, call Buildkite API to start a new build
+      buildkite_trigger_build(
+        buildkite_organization: BUILDKITE_ORG_NAME,
+        buildkite_pipeline: BUILDKITE_PIPELINE,
+        **common_args
+      )
     end
   end
 
   #####################################################################################
-  # Version Methods
+  # Versioning utils
   #####################################################################################
 
   # Updates the build code in version.properties to the next build code and commits the change
@@ -132,7 +186,7 @@ platform :android do
       version_name: release_version_current,
       version_code: build_code_next
     )
-    commit_version_bump
+    commit_version_update(message: '[skip ci] Bump build code')
   end
 
   # Updates the version name in version.properties to the next release version and commits the change
@@ -141,14 +195,14 @@ platform :android do
     VERSION_FILE.write_version(
       version_name: release_version_next
     )
-    commit_version_bump
+    commit_version_update(message: '[skip ci] Bump app version')
   end
 
-  # Commits the version bump to the version.properties file
+  # Commits the version update to the version.properties file
   #
-  def commit_version_bump
+  def commit_version_update(message:)
     Fastlane::Helper::GitHelper.commit(
-      message: 'Bump version number',
+      message: message,
       files: VERSION_PROPERTIES_PATH
     )
   end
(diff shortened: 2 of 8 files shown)


fastlane/Fastfile
```diff
@@ -61,19 +110,166 @@ platform :android do
     )
   end
 
-  def generate_prototype_build_number
-    if ENV['BUILDKITE']
-      commit = ENV.fetch('BUILDKITE_COMMIT', nil)[0, 7]
-      branch = ENV['BUILDKITE_BRANCH'].parameterize
-      pr_num = ENV.fetch('BUILDKITE_PULL_REQUEST', nil)
+  lane :build_and_upload_to_play_store do |track: 'internal', release_status: 'draft'|
+    # TODO using `internal`` track with `draft` releases by default while the release itself is in draft mode
```
dangermattic, line 114: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/Gravatar-Android/blob/70618e4fbb7fe6a73f6f71fc7121820c094d3033/fastlane/Fastfile#L114"></span>Style/CommentAnnotation: Annotation keywords like <code>TODO</code> should be all upper case, followed by a colon, and a space, then a note describing the problem.</td>
    </tr>
  </tbody>
</table>

.buildkite/release-pipelines/new-alpha-release.yml
```diff
@@ -0,0 +1,18 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+---
+
+steps:
```
iangmaia, line 4: This pipeline is to start the alpha release process completely from CI (eventually from ReleasesV2).

hamorillo (approved):
I can't say much about the pipeline code 😅 . I've tested it and works as expected. I can see the release on Google Play Console:

image

fastlane/Fastfile
```diff
@@ -7,19 +7,68 @@ UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Hel
...
+    require_env_vars!('PLAY_STORE_RAW_JSON_KEY')
+
+    upload_to_play_store(
+      package_name: 'com.gravatar.app',
+      aab: File.join(PROJECT_ROOT_FOLDER, 'app', 'build', 'outputs', 'bundle', 'release', 'app-release.aab'),
+      track: track,
+      release_status: release_status,
+      rollout: rollout,
+      skip_upload_metadata: (track != 'production'),
+      skip_upload_changelogs: true,
```
AliSoftware: I know this PR is only initial setup and you haven't test/integrated the flow for releasing to production track just yet, but since you used such a condition for `skip_upload_metadata` I think it'd make sense to do the same for `skip_upload_changelogs` too, since every time we submit to Production we'd want the `changelogs/*.txt` files to be used to provide release notes for the release (just like we'd like to ensure the Play Store Listing metadata would be up-to-date)?

```suggestion
      skip_upload_changelogs: (track != 'production',
```

fastlane/Fastfile
```diff
@@ -7,19 +7,68 @@ UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Hel
...
+      skip_upload_images: true,
+      skip_upload_screenshots: true,
+      json_key_data: ENV.fetch('PLAY_STORE_RAW_JSON_KEY', nil)
```
AliSoftware: ❤️  I like that we now provide this key via a multiline env var 👍  rather than via a file installed via `configure_apply` like is the case for legacy reasons in WC or WP for example (where the key is stored in a `.json` file installed via `configure` and their Fastfile then use `json_key: <path to decrypted file>`). Using env var + `json_key_data` is indeed the new way to do it in our infra (and at some point we should probably convert other projects to use the env var like here instead too) 👍 

💭 I see that in `~/.mobile-secrets` you've injected that env var directly in the `env` file for the pipeline though (and having to resort to the `=$(cat <<'EOF'…)` trick).
💡 I think it'd make more sense to put that key in the `shared-secrets.yml` file of `mobile-secrets`, and then just reference it in `CI/secrets/gravatar-android/env`:
 - This way, if we continue adding more Android projects later, or if we finally find the time one day to migrate existing Android projects like WC and WP to use `json_key_data` and env var instead of `json_key` + `configure`, we'll already be able to reuse that same secret in more project without duplicating it in every place
 - But also so that you'd be able to benefit from the fact that `${…}` substitutions that are post-processes on `env` files now support multiline strings from `shared-secrets.yml` and handle them automatically for you, without you have to resort to the `=$(cat <<'EOF'…)` trick explicitly nor manually

(See paaHJt-7Pd-p2#usage for details)

fastlane/Fastfile
```diff
@@ -61,19 +110,166 @@ platform :android do
     )
   end
 
-  def generate_prototype_build_number
-    if ENV['BUILDKITE']
-      commit = ENV.fetch('BUILDKITE_COMMIT', nil)[0, 7]
-      branch = ENV['BUILDKITE_BRANCH'].parameterize
-      pr_num = ENV.fetch('BUILDKITE_PULL_REQUEST', nil)
+  lane :build_and_upload_to_play_store do |track: 'internal', release_status: 'draft'|
+    # TODO: using `internal`` track with `draft` releases by default while the release itself is in draft mode
```
AliSoftware:
```suggestion
    # TODO: using `internal` track with `draft` releases by default while the release itself is in draft mode
```

fastlane/Fastfile
```diff
@@ -61,19 +110,166 @@ platform :android do
     )
   end
 
-  def generate_prototype_build_number
-    if ENV['BUILDKITE']
-      commit = ENV.fetch('BUILDKITE_COMMIT', nil)[0, 7]
-      branch = ENV['BUILDKITE_BRANCH'].parameterize
-      pr_num = ENV.fetch('BUILDKITE_PULL_REQUEST', nil)
+  lane :build_and_upload_to_play_store do |track: 'internal', release_status: 'draft'|
```
AliSoftware, line 133: nit: we might want to add some YARD doc to all those new lanes especially to document each `@param` type and role

fastlane/Fastfile
```diff
@@ -61,19 +110,166 @@ platform :android do
...
+      This will:
+      - Checkout and pull the #{DEFAULT_BRANCH} branch (if running on CI)
+      - Bump the build code from #{build_code_current} to #{build_code_next}
+      - Trigger a new alpha build in Buildkite
+    PROMPT
+    next unless skip_confirm || UI.confirm('Continue?')
+
+    Fastlane::Helper::GitHelper.checkout_and_pull(DEFAULT_BRANCH) if is_ci?
+
+    bump_build_code
```
AliSoftware, line 165: I know that we are only doing alphas for this initial setup, but just thinking ahead that if we later plan to start doing both alphas and betas, I'm not 100% sure this will work well?

For example, if you do the beta first, which will bump the version on `release/*`), then do the alpha (manually or because it'd be on a Buildkite schedule) after that (which will bump the version on `trunk`… so potentially to the same version as what was just bumped on `release/*`) before doing the backmerge, you'll end up trying to upload the same `versionCode` twice.

Actually even if you do the alpha first it might also be worse, as the `release/*` will never see (never receive the commit via backmerge) the commit that did the version bump on `trunk` for the alpha, so in that case the conflicting `versionCode` case will happen for sure.

---

For that reason, we might want to adopt a logic for the `versionCode` that is derived from `versionName` (like we do on Tumblr), e.g. for a `versionName` of `x.y.z-rc-N` then use `XXYYZZNN` as `versionCode`. You should be able to use something like [the `DerivedBuildCodeFormatter`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/versioning/formatters/derived_build_code_formatter.rb) class for that (which also means you won't need the `BuildCodeCalculator` anymore when using a derived build code)

This approach has many benefits:
 - It's more deterministic and thus easier to map and reason about
 - Won't ever risk causing issue with conflicting bumps from alpha vs beta `versionCodes`
 - But more importantly, it won't require you to have to do a new beta when you need to do a hotfix, and will reduce the issues with the exceptions we have to go through in versionning hotfixes for other apps. For example if the current version in code freeze is `1.3.0` (`versionCode=01030005` if, say, it's the 5th build of that version), and you need to do a hotfix for `1.2.1`, then its `versionCode=01020101` will be guaranteed to be smaller than any `versionCode` used by `1.3.0` and thus only served as an update to users that were still on `1.2.0` but not to the ones already on `1.3.0`.

We've been willing to adopt such a versioning system for a while on other apps to avoid that age-old problem once and for all, but on an app that has already been on the Play Store for a while and has already consumed some `versionCode` in its lifetime (+ the fact that there's a max value for `versionCode` allowed by Play Store) makes it tricky for existing apps (hence why we still haven't tacked this).

But for a brand new app like Gravatar Android, it might be the ideal occasion to start doing that right from the get go 🙂

fastlane/Fastfile
```diff
@@ -7,19 +7,68 @@ UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Hel
...
+    require_env_vars!('PLAY_STORE_RAW_JSON_KEY')
+
+    upload_to_play_store(
+      package_name: 'com.gravatar.app',
+      aab: File.join(PROJECT_ROOT_FOLDER, 'app', 'build', 'outputs', 'bundle', 'release', 'app-release.aab'),
+      track: track,
+      release_status: release_status,
+      rollout: rollout,
+      skip_upload_metadata: (track != 'production'),
+      skip_upload_changelogs: (track != 'production',
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/Gravatar-Android/blob/89e438fab938982960cfc8461705700e54342ace/fastlane/Fastfile#L57"></span>Lint/Syntax: unexpected token tCOMMA<br />
(Using Ruby 3.2 parser; configure using <code>TargetRubyVersion</code> parameter, under <code>AllCops</code>)</td>
    </tr>
  </tbody>
</table>

fastlane/Fastfile
```diff
@@ -7,19 +7,68 @@ UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Hel
...
+      skip_upload_images: true,
```
dangermattic, line 58: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/Gravatar-Android/blob/89e438fab938982960cfc8461705700e54342ace/fastlane/Fastfile#L58"></span>Lint/Syntax: cannot assign to a keyword<br />
(Using Ruby 3.2 parser; configure using <code>TargetRubyVersion</code> parameter, under <code>AllCops</code>)</td>
    </tr>
  </tbody>
</table>

fastlane/Fastfile
iangmaia: Love it! Yeah that way we can indeed make other apps to share the same reference in `shared-secrets.yml`.

I've decided to change the var name in `mobile-secrets`, so I updated it on a0487a39acf82c9d0e1a824b02c24036d57d9969.

fastlane/Fastfile
iangmaia: Love it (2)! I don't think any other project other than Tumblr uses that pattern now? I'll update the versioning to do it. I agree it's a nice pattern to start with.

AliSoftware: PS: I wonder if it's worth adding the `1` prefix like `DerivedBuildCodeFormatter` currently does in all cases, especially given the max value of `versionCode` allowed by the Play Store Console is ≈ `2_000_000_000` and having the `1` prefix in addition of the 4 * 2 digits (major, minor, patch, rc) makes it closer to that limit (`1XX_YYZ_ZNN`),

So might be worth updating the `release-toolkit`'s `DerivedBuildCodeFormatter` to have the `prefix` as an instance variable of that class, add a `def initialize(prefix:)` constructor that sets it, and use it during the call to `def build_code`.

That way when you set `BUILD_CODE_FORMATTER = Fastlane::Wpmreleasetoolkit::Versioning:DerivedBuildCodeFormatter.new(prefix: '')` that allows you to not have a prefix if you don't think you'll need it, while still allowing potential other products that want the `'1'` prefix to keep it… or provide a different prefix like `2` (for wear or Android auto variants for example)

Note that for this to work well you'd need to trim any potential leading zero from the result (e.g. `.gsub(/^0+/, '')`, so that if the major version (`XX`) is < 10 and the `prefix = ''`, the `def build_code` would not include a leading zero 😉 

[EDIT] Could also be worth checking if the `prefix` is valid in `initialize` and `UI.user_error!` if it's not an integer, or if its length is > 1 (or if it's `to_i` value is > 20… but at that point I think it's better not to encourage flying too close to the ~~sun~~ max value by allowing anything even >9 imho)

fastlane/Fastfile
iangmaia: Updated on d3fca8135f3c34e80c096dc825bac2c106580341

fastlane/Fastfile
iangmaia: Added the usage of `DerivedBuildCodeFormatter` (as it is, for now) on 89a5d401cb8cfa9205baa9b3859bee09aeeb9676.

AliSoftware: Errrr be sure not to run any lane using it then and not push any build to Google Play while it's used in its current form, otherwise you risk pushing a build with a very high `versionCode` (due to `1` extra prefix) that would suddenly set the minimum value for all subsequent `versionCode` for this app going forward with no way of going back to a lower `versionCode` without prefix… 😓

iangmaia: 😮‍💨 oof
On Friday I wanted to test the submission to alpha tracks straight away (which, looking on the bright side, worked well after all the info that was blocking the release was added) and submitted it. So indeed we lost ~100M versions (starting from `100020001`) out of a [maximum](https://support.google.com/googleplay/android-developer/answer/9859152?hl=en#zippy=%2Capp-version-requirements-for-play-console) of ~2BI (`2100000000`) 😕 😞 .

Imagining we follow this pattern though (now with the prefix anyway 🙄), it should still be ok? Our pattern itself gives us 10k `major.minor` combinations (assuming `major`/`minor` versions `0`-`99`) which is probably more than enough.
Another change I thought just now in the `DerivedBuildCodeFormatter` would be to allow customizing the number of different digits per component so, for example, we could use 3 digits for minor versions and patch only 1 digit (or use one extra digit and assign 3 digits for any of the components at any time).

Anyway, I've created a PR for making the prefix optional, though we won't be able to use it here: https://github.com/wordpress-mobile/release-toolkit/pull/656.

AliSoftware: > On Friday I wanted to test the submission to alpha tracks straight away (which, looking on the bright side, worked well after all the info that was blocking the release was added) and submitted it. So indeed we lost ~100M versions (starting from 100020001) out of a [maximum](https://support.google.com/googleplay/android-developer/answer/9859152?hl=en#zippy=%2Capp-version-requirements-for-play-console) of ~2BI (2100000000)

I think maybe if you remove all the offending `.aab`/`.apk` from the Google Play Console ("App Bundle Explorer") we might be able to upload builds with lower `versionCode` again after all… as long as those builds never got submitted for review and published to end users? Which would allow us to get rid of the prefix and reclaim those lost ≈100M version codes 😓 

> Imagining we follow this pattern though (now with the prefix anyway 🙄), it should still be ok? Our pattern itself gives us 10k major.minor combinations (assuming major/minor versions 0-99) which is probably more than enough.

True, but we also never know what the future holds, and that also means we're locked into our version format too now… while it's better to leave those doors open especially given how restrictive it'd be (i.e. no ability to push any new version to Google Play if we reached the max, so the farther we're from that limitation the better given such a limitation would prevent us from ever submitting any new version if we reached it, which would be quite critical)

> Another change I thought just now in the DerivedBuildCodeFormatter would be to allow customizing the number of different digits per component so, for example, we could use 3 digits for minor versions and patch only 1 digit (or use one extra digit and assign 3 digits for any of the components at any time).

Great idea 👍  I like it

iangmaia: > we also never know what the future holds

> the farther we're from that limitation the better given such a limitation would prevent us from ever submitting any new version if we reached it, which would be quite critical

True 👍

> I think maybe if you remove all the offending .aab/.apk from the Google Play Console ("App Bundle Explorer") we might be able to upload builds with lower versionCode again after all

Yeah, that's what I was trying to search for today and couldn't find. Maybe if the release was still a draft it would be possible, but once it gets distributed to users (even on internal tracks, apparently?) it gets "written in stone" 😢 
A step before was that not only the release but the whole app was in draft mode, so it took us some time to have all the data filled to make it move out of that state. Then I wanted to test if everything was correct and submitted the draft for internal testing... 😕 

Anyway. I've added version components digits configuration in release-toolkit in https://github.com/wordpress-mobile/release-toolkit/pull/657, so perhaps we can use that here foreseeing more room for releases 🤔 (suggestions are welcome 🙂)

iangmaia:
@AliSoftware  I've updated release-toolkit to `13.4.0` and the versioning scheme in the project.
I'll run tomorrow a `new_alpha_release` -- we'll need to first bump the `versionName` to `0.21`, given it now the `minor` component has 3 digits, but from what I see everything seems fine in the new `DerivedBuildCodeFormatter`. Then we can finally merge this PR and have this versioning moving forward.

fastlane/Fastfile
```diff
@@ -7,19 +7,78 @@ UI.user_error!('Please run fastlane via `bundle exec`') unless FastlaneCore::Hel
...
+
+PROJECT_ROOT_FOLDER = File.dirname(File.expand_path(__dir__))
+
+VERSION_PROPERTIES_PATH = File.join(PROJECT_ROOT_FOLDER, 'version.properties')
+VERSION_FILE = Fastlane::Wpmreleasetoolkit::Versioning::AndroidVersionFile.new(version_properties_path: VERSION_PROPERTIES_PATH)
+
+VERSION_CALCULATOR = Fastlane::Wpmreleasetoolkit::Versioning::SemanticVersionCalculator.new
+VERSION_FORMATTER = Fastlane::Wpmreleasetoolkit::Versioning::RCNotationVersionFormatter.new
+# Using DerivedBuildCodeFormatter to derive build codes from version name (format: 1XXYYYZNN)
+# This eliminates the need for manual build code incrementing and provides deterministic versioning
```
AliSoftware: 💡 / 💄 
```suggestion
# This eliminates the need for manual build code incrementing, provides deterministic versioning,
# and also avoids the need to do a new beta for version N+1 every time we do a hotfix for version N
```

fastlane/Fastfile
```diff
@@ -61,19 +120,192 @@ platform :android do
...
+#
+def build_code_next
+  current_version = VERSION_FORMATTER.parse(version_name_current)
+  # Create a new version with incremented build number
+  next_version = Fastlane::Models::AppVersion.new(
+    current_version.major,
+    current_version.minor,
+    current_version.patch,
+    current_version.build_number + 1
+  )
```
AliSoftware, line 282: 💡 🔜  Note for the future (definitively not for this PR, just ideas for improvement later):

1️⃣  We have `*BuildCodeCalculator` classes in `release-toolkit` (e.g. [`SimpleBuildCodeCalculator`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/versioning/calculators/simple_build_code_calculator.rb), [`DateBuildCodeCalculator`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/versioning/calculators/date_build_code_calculator.rb)) to compute the next build code for those.
So we should probably consider creating a similar `DerivedBuildCodeCalculator` in `release-toolkit`, with `def next_build_code(version:)` method that would do what you did here, for us to abstract that logic away from the `Fastfile`.

2️⃣  It has also been in my mind for a very long time to also make a `VersionManager` class that would expose all those `build_code_current`, `build_code_next`, `release_version_current`, `release_version_next` methods that we keep redeclaring in the `Fastfile` of all of our repos, and that would encapsulate the assembly of the version calculator + formatter and build code calculator + formatter, so that we can still assemble each individual component like pieces of lego, but in the end have a single object to manipulate. For example:

```ruby
VERSION_MANAGER = Fastlane::Wpmreleasetoolkit::Versioning::VersionManager.new(
  version_calculator:  Fastlane::Wpmreleasetoolkit::Versioning::SemanticVersionCalculator.new,
  version_formatter: Fastlane::Wpmreleasetoolkit::Versioning::RCNotationVersionFormatter.new,
  build_code_calculator:  Fastlane::Wpmreleasetoolkit::Versioning::DerivedBuildCodeCalculator.new,
  build_code_formatter: Fastlane::Wpmreleasetoolkit::Versioning::DerivedBuildCodeFormatter.new(prefix: '1', major_digits: 2, minor_digits: 3, patch_digits: 1, build_digits: 2)
)
```
And then we could do every version-related actions in the rest of the `Fastfile` using the `VERSION_MANAGER` as the single API interface to interact with, which would take care to delegate to the calculators/formatters as needed.
For example, just calling `VERSION_MANAGER.build_code_next` would take care of using its `@version_formatter` to parse the version file, use its `@build_code_calculator` to increment the value, and its `@build_code_formatter` to return the incremented build code—mostly like you do here in your own `def build_code_next` method, except that we wouldn't need to declare such a method in each of our `Fastfile` files anymore as the implementation would have been DRY'ed into the `VersionManager` class from `release-tookit`.

To get there we might need to first make sure all the `*BuildCode*` classes inherit from the same abstract class and have a common API, so that the `VersionManager` class can call the same methods on those regardless of which subclass is actually used. But I think the end result would be really great with a lot of simplification in our Fastfiles and all more consistent APIs and setups across repos.

Of course, this would be its own project, for a later time, and completely unrelated to this PR. But I figured I'd share the idea while at it, to keep in the back of our minds.

AliSoftware (approved):
:shipit:

fastlane/Fastfile
iangmaia: > make a VersionManager class that would expose all those build_code_current, build_code_next, release_version_current, release_version_next methods that we keep redeclaring in the Fastfile of all of our repos

As @ParaskP7 says, 💯 ^ 💯 ^ 💯 ! 😄 

Something like that also crossed my mind multiple times. This is pretty much a constant across many repos and it would be great to also standardize on naming (some repos have these common functions slightly different).

ParaskP7: > As @ParaskP7 says, 💯 ^ 💯 ^ 💯 ! 😄

Ah, I am just using 💯 x 💯 ^ 💯 , using `x` then `^` 😄 but you took it to the next level, using `^` followed by another `^` 🐡 my brain just can't cope with this yet! 🤣 🤣 🤣

PR #87, opened 2025-07-28
[Tooling] Remove GitHub check about Prototype Build

iangmaia:
### Description

Follow-up on the brief discussion we had [here](https://github.com/Automattic/Gravatar-Android/pull/65#issuecomment-3121221882) about the Prototype Build GitHub check.

@hamorillo mentioned the Prototype would still always show as pending, even if they don't plan running it, which can be confusing.

The main downside of not showing it at all is making it even more "buried", but again it's about making a choice.
We have a few options:
1. Always run the Prototype Build
2. Optionally run the Prototype Build, with the GitHub check about the build showing as pending
3. Optionally run the Prototype Build but with no pending check (looks nicer but we might "forget" about prototype builds?)
4. Optionally run the Prototype Build with no pending check, but show a message (perhaps with a link to Buildkite) using Danger saying that we can run the Prototype Build if we want (with the downside of always having this message in PRs)

I'm leaving the options here and personally have no strong opinions -- I'd be fine with 3 (meaning merging this PR) if the team decides to do so.

### Testing Steps

The Prototype Build GitHub check shouldn't show.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -97,6 +97,3 @@ steps:
     plugins: [$CI_TOOLKIT]
     artifact_paths:
       - "**/build/outputs/apk/**/*"
-    notify:
-      - github_commit_status:
-          context: Prototype Build
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -91,6 +91,7 @@ steps:
     blocked_state: passed
 
   - label: 🏗️ Prototype Build
+    if: build.pull_request.id != null
     depends_on:
       - run-on-demand-prototype-build
     command: .buildkite/commands/prototype-build.sh


AliSoftware:
Technical changes in the pipeline looks good to me, but I'm gonna leave the approval decision to the Gravatar Android dev team as to which of the 4 options you listed they actually prefer.

hamorillo (approved):
I would go with option 3 or 4. Let's continue with the current approach, and we can add the comment later if we think it will be helpful.

Thanks for fixing this.

AliSoftware:
💡  Option 4b: include a reminder about how one can generate a Prototype Build in the [`.github/PULL_REQUEST_TEMPLATE.md`](https://github.com/Automattic/Gravatar-Android/blob/trunk/.github/PULL_REQUEST_TEMPLATE.md).

That way there would still be a reminder, but not one generated by a comment that would create an additional notification and noise of every PR. Yet the reminder would still be included in every PR (assuming PR author doesn't delete that part of the template)

pinarol:
Planning to make a call for testing for Android so we'll need the prototype builds and we need to revert this PR for that. We'll take care of the revert PR.

AliSoftware:
> Planning to make a call for testing for Android so we'll need the prototype builds and we need to revert this PR for that. We'll take care of the revert PR.

@pinarol  I'm not sure I'm following… the whole goal of this PR was to make the Prototype Builds still be available, but only on demand, and not being marked as "Pending" on GitHub PR checks if a prototype build was not requested. But you should still be able to trigger Prototype Builds on each of your PRs when needed, right? At least I thought that was the whole point.

Actually, your comment above might make my point on the Prototype Builds not being reported nor advertised as being possible (yet optional) anywhere, making it not clear enough (1) that those are still available (2) what the process is to generate a Prototype build (3) making it easily forgettable that those are a thing.

So maybe we should implement option 4a (generate a PR comment with instructions on how to trigger a Prototype Build on demand on the PR if wanted) or 4b (update the `PULL_REQUEST.md` template to include those instructions (that will hopefully be kept and not be deleted when people write their PR descriptions)?

iangmaia:
I've reverted the reversion on https://github.com/Automattic/Gravatar-Android/pull/101

> Actually, your comment above might make my point on the Prototype Builds not being reported nor advertised as being possible (yet optional) anywhere, making it not clear enough (1) that those are still available (2) what the process is to generate a Prototype build (3) making it easily forgettable that those are a thing.
> 
> So maybe we should implement option 4a (generate a PR comment with instructions on how to trigger a Prototype Build on demand on the PR if wanted) or 4b (update the `PULL_REQUEST.md` template to include those instructions (that will hopefully be kept and not be deleted when people write their PR descriptions)?

Yep, I thought about it. As soon as new developers landed in the codebase, the prototypes were hidden 😓 Doesn't happen very often but likely will happen again...

PR #88, opened 2025-07-29
Open TermsOfService and PrivacyPolicy in a customTab

hamorillo:
### Description

`TermOfService` and `PrivacyPolicy` links should be opened in an in-app browser so the user doesn't leave the app context.

https://github.com/user-attachments/assets/d81d7778-7ee4-4cdf-a00b-70d321e32d32

### Testing Steps

- Launch de app
- Open the `About this app` dialog
- Verify `TermOfService` and `PrivacyPolicy` are opened in an in-app browser.


etoledom (approved):
Working perfectly as described 🎉

PR #89, opened 2025-07-29
Improve blurred header look

etoledom:
Closes GRA-631

### Description

This PR modifies the parameters on the Blurred Header Background for the blur effect and color to look closer to the designs and iOS.

I also took the chance to modify the labels in the profile header. Making them the same size and color as in the designs.

Previously:
CleanShot 2025-07-29 at 14 26 23@2x

Now:
CleanShot 2025-07-29 at 14 25 53@2x
CleanShot 2025-07-29 at 14 25 06@2x
CleanShot 2025-07-29 at 14 24 39@2x

### Testing Steps

- Run the app
- Check that the header looks closer to the designs.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/header/DisplayName.kt
```diff
@@ -15,12 +16,12 @@ internal fun DisplayName(displayName: String, modifier: Modifier = Modifier) {
         text = displayName,
         style = MaterialTheme.typography.titleMedium.copy(
             fontWeight = FontWeight.Bold,
-            color = MaterialTheme.colorScheme.onSurface,
+            color = Color.White,
```
etoledom: Not sure if this is the best way to do this.

On iOS we can set a specific component to `.environment(\.colorScheme, .dark)` .
I couldn't find a simile on Jetpack Compose.

Still this should be always white, and is how the link text color is set also.

hamorillo: I think it would be better to set the `AnimatedProfileHeaderSavedState` to always use the darkTheme. I've done the changes [here](https://github.com/Automattic/Gravatar-Android/tree/hamorillo/blurred-header-look-suggestion), so it's easier to see what I mean.

Diff: https://github.com/Automattic/Gravatar-Android/compare/etoledom/blurred-header-look...hamorillo/blurred-header-look-suggestion

If you like it, you can cherry-pick the last commit.

_Note:_ I've also regenerated the screenshots in that commit.

hamorillo (commented):
Thanks for the improvements! Looks great! I've left a comment about the color approach. :)

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/header/DisplayName.kt
etoledom: I see! So `GravatarAppTheme(darkTheme = true)` is the trick 👍

etoledom:
@hamorillo - I've merged https://github.com/Automattic/Gravatar-Android/tree/hamorillo/blurred-header-look-suggestion into this branch 👍

hamorillo (approved):
Looks great! Thanks for taking care of this! 👨‍🎨 🚀

PR #91, opened 2025-07-29
Adding Location label to the profile header

etoledom:
### Description

This PR adds the location label to the profile header.

https://github.com/user-attachments/assets/2e84cf8e-542d-4b4c-a538-f42550f27350

### Testing Steps

- Check the location label on the Profile header.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/header/LocationInfo.kt
```diff
@@ -0,0 +1,25 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.sp
+
+@Composable
+internal fun LocationInfo(locationInfo: String, modifier: Modifier = Modifier) {
+    BasicText(
+        text = locationInfo,
+        style = MaterialTheme.typography.bodyMedium.copy(
+            color = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.8f)
```
hamorillo, line 16: ⚠️ As per designs, the color should be the same as in `JobInfo`.

```suggestion
            color = MaterialTheme.colorScheme.onSurface.copy(alpha = 0.6f)
```

hamorillo (commented):
💡`JobInfo` and `LocationInfo` should be identical based on the design, so maybe we can remove the `LocationInfo` created in this PR and rename the `JobInfo` to something more generic. (Ex: ProfileInfo). With this approach, if we need to update the color or any other properties, we will need to update only one place. Wdyt?

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/profile/header/LocationInfo.kt
etoledom: We changed the JobInfo color to alpha 0.8 on https://github.com/Automattic/Gravatar-Android/pull/89

So I thought to give it 0.8 right away to not change it later again.

etoledom:
@hamorillo - I have updated from trunk after merging https://github.com/Automattic/Gravatar-Android/pull/89

Now JobInfo and LocationInfo have the same text color, as shown in the updated snapshot 👍

hamorillo (approved):
Looks good! 🚀

PR #92, opened 2025-07-30
Pull to refresh on avatars grid will now refresh avatars in the header

etoledom:
Closes GRA-640

### Description

If the user has changed their selected avatar in another client, pull to refresh should reflect this change also in the avatars on the header, not just on the grid.

https://github.com/user-attachments/assets/5e1534de-7628-432f-9e9d-bec0b070341a

### Testing Steps

- Run the app
- Go to a different client and change the selected avatar
- On this build, go to the Avatar Picker screen 
- Do a Pull to refresh.
  - Check that the grid has updated selecting the current selected avatar
  - Check that the header has updated showing the current selected avatar

### DIFF
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
@@ -1,0 +1,27 @@
+package com.gravatar.app.usercomponent.domain.usecase
+
+import com.gravatar.app.usercomponent.data.AvatarCacheBusterStorage
+import com.gravatar.app.usercomponent.domain.repository.UserRepository
+import com.gravatar.restapi.models.Avatar
+
+internal class FetchAvatarsUseCase(
+    private val userRepository: UserRepository,
+    private val avatarCacheBusterStorage: AvatarCacheBusterStorage,
+) : FetchUserAvatars {
+
+    override suspend fun invoke(isRefreshing: Boolean): Result<List<Avatar>> {
+        return userRepository.getAvatars()
+            .onSuccess { avatars ->
+                if (isRefreshing) {
+                    val avatarID = avatars.firstOrNull { it.selected == true }?.imageId
+                    avatarID?.let { id ->
+                        avatarCacheBusterStorage.saveAvatarCacheBuster(id)
+                    }
+                }
+            }
+    }
+}
+
+interface FetchUserAvatars {
+    suspend operator fun invoke(isRefreshing: Boolean): Result<List<Avatar>>
+}
--- a/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
+++ b/userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
@@ -7,21 +7,20 @@ import com.gravatar.restapi.models.Avatar
 internal class FetchAvatarsUseCase(
     private val userRepository: UserRepository,
     private val avatarCacheBusterStorage: AvatarCacheBusterStorage,
+    private val clock: AppClock,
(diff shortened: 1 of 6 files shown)


userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
```diff
@@ -0,0 +1,27 @@
...
+        return userRepository.getAvatars()
+            .onSuccess { avatars ->
+                if (isRefreshing) {
+                    val avatarID = avatars.firstOrNull { it.selected == true }?.imageId
+                    avatarID?.let { id ->
+                        avatarCacheBusterStorage.saveAvatarCacheBuster(id)
+                    }
+                }
+            }
+    }
```
etoledom: I have copied the mechanism other requests use to force-refresh the avatar. I hope it's the correct one in this case also.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
```diff
@@ -0,0 +1,27 @@
...
+}
+
+interface FetchUserAvatars {
+    suspend operator fun invoke(isRefreshing: Boolean): Result<List<Avatar>>
+}
```
hamorillo: ⚠️ We should update the cacheBuster not only when pulling to refresh. We need to update the cacheBuster every time we fetch the avatars. The current implementation doesn't solve the case where I have my app closed, and modify the selected avatar on a different device:

https://github.com/user-attachments/assets/a43fbee5-0e20-4a5c-ab4d-a2fb858d7018

We should also consider the case where the avatar has been deleted on a different device, so I think the code should look something like:

```suggestion
internal class FetchAvatarsUseCase(
    private val userRepository: UserRepository,
    private val avatarCacheBusterStorage: AvatarCacheBusterStorage,
    private val clock: AppClock,
) : FetchUserAvatars {

    override suspend fun invoke(): Result<List<Avatar>> {
        return userRepository.getAvatars()
            .onSuccess { avatars ->
                val selectedAvatarId = avatars.firstOrNull { it.selected == true }?.imageId
                avatarCacheBusterStorage.saveAvatarCacheBuster(
                    selectedAvatarId ?: clock.now().toString()
                )
            }
    }
}

interface FetchUserAvatars {
    suspend operator fun invoke(): Result<List<Avatar>>
}
```

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
```diff
@@ -0,0 +1,27 @@
+package com.gravatar.app.usercomponent.domain.usecase
+
+import com.gravatar.app.usercomponent.data.AvatarCacheBusterStorage
+import com.gravatar.app.usercomponent.domain.repository.UserRepository
+import com.gravatar.restapi.models.Avatar
+
+internal class FetchAvatarsUseCase(
```
hamorillo, line 8: 💡 It would be nice to add some unit test to this UseCase

hamorillo (commented):
Thanks for taking care of this @etoledom! I left some comments about the current approach.

userComponent/src/main/kotlin/com/gravatar/app/usercomponent/domain/usecase/FetchAvatarsUseCase.kt
etoledom: Makes total sense! Thank you for the suggestions and changes 🙏

etoledom:
@hamorillo - I have committed the proposed change, made the corresponding updates, and added a test suite for `FetchAvatarsUseCase` 👍 

Now it also works updating the header avatars when the app starts if needed.

Ready for another look 👀

hamorillo (approved):
Thanks for the changes! Works great! 🚀

PR #93, opened 2025-07-30
Implement Share vCard feature

hamorillo:
### Description

This PR implements the Share vCard option in the Share Screen. In this case, when the user shares their contact, we want to include in the `.VCF` the currently selected avatar.

The Gravatar App should open the system's share sheet showing all the apps that accept that file.

image

For the file name, we are using the `Display Name`.

Different apps handle that file in different ways. Some messaging apps parse the VCF and present it in a concrete way. For example, WhatsApp presents the contact in the conversation, but as we are not including the `FN`, they use the `ORG` instead of `Nickname` or `N`. (cc: @pinarol)

image

When we try to add the contact to the agenda from those apps, sometimes we lose the photo or other data. I don't think we can do much about that :( 
However, when adding the contact using the file, we can see the avatar:

image

### Testing Steps

- Launch the app
- Switch to the Share Screen
- Click into the Share icon
- Verify the system share sheet is opened
- Try to share with different apps (messaging, email, etc.)
- Save the file locally (for example, sharing the .vcf file with a file explorer)
- Using the same file explorer, open the file and verify it's imported.

**Note:** As each app handles the VCF file differently, please feel free to comment on unexpected behaviours and we'll see what we can do (if any).


etoledom (approved):
Looking great!

Tested sharing over WhatsApp and Signal. Though is very little info what WhatsApp parses, its out of our control. Similar case with Signal.

I also tested opening the shared .vcf file (via email) and the avatar is there as expected 👍 

**Note:** Probably not related to this PR. I noticed that the location is not being shared neither via QR nor Share option.

etoledom:
> Note: Probably not related to this PR. I noticed that the location is not being shared neither via QR nor Share option.

Added via https://github.com/Automattic/Gravatar-Android/pull/96

hamorillo:
> Note: Probably not related to this PR. I noticed that the location is not being shared neither via QR nor Share option.

Good catch! Thanks

PR #94, opened 2025-07-30
Remove "make current" option on selected avatars

etoledom:
Closes GRA-642

### Description

This PR removes the "Make current" option on the already selected avatar.

CleanShot 2025-07-30 at 19 45 48@2x
CleanShot 2025-07-30 at 19 46 06@2x

### Testing Steps

- Tap on a non-selected avatar
  - Check that the "make current" option is there
- Tap on the currently selected avatar
  - Check that the "make current" option is not there

### DIFF
--- a/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/AvatarMoreOptionsPickerPopup.kt
+++ b/homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/AvatarMoreOptionsPickerPopup.kt
@@ -19,14 +19,14 @@ import com.gravatar.ui.GravatarTheme
 
 @Composable
 internal fun AvatarMoreOptionsPickerPopup(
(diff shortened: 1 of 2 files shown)


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/AvatarMoreOptionsPickerPopup.kt
```diff
@@ -19,17 +19,15 @@ import com.gravatar.ui.GravatarTheme
 
 @Composable
 internal fun AvatarMoreOptionsPickerPopup(
+    isAvatarSelected: Boolean,
```
hamorillo: ⛏️ _Totally nitpicking:_ I would rename this to `isSelected` because we already know we are talking about the avatar. But feel free to omit this comment.

hamorillo (approved):
🚀  Great improvement! Works as described.

PR #96, opened 2025-07-31
Add location to VCard

etoledom:
Closes GRA-647

### Description

This PR adds the location field to the VCard, to be shared via QR and Share sheet.

### Testing Steps

- Switch ON sharing location
- Scan the QR code
  - Check that the location is shared.


hamorillo (approved):
🚀

PR #97, opened 2025-07-31
Implement Expanded QR

hamorillo:
### Description

This PR implements the expanded QR code. In the Share Screen, we are adding a new option that allows users to display the QR Code with the contact information in a full-screen presentation.

https://github.com/user-attachments/assets/32ff34f4-90e2-4631-8b43-67e35e6bea12

### Testing Steps

- Launch the app and switch to the share screen
- Click on the expand icon
- Verify the QrCode is presented in full screen (system bars should disappear)
- Tap the close button and verify you navigate to the Share screen (system bars should appear)
- Click on the expand icon
- Verify the QrCode is presented in full screen (system bars should disappear)
- Press the back button and verify you navigate to the Share screen (system bars should appear)


etoledom (commented):
Looking great!

Just a few comments (only the first one requires change):

### Could we give the close button a bit of padding at the top? 
It feels like it's way too close to the top of the screen, and it momentary gets under the status bar.

Maybe something like this (even if it's not exact):
CleanShot 2025-08-01 at 11 55 56@2x


### The orientation of the presentation was a bit unexpected

The view presenting from the top left corner felt unexpected 🤔 
If the animation could come from straight up-down, or down-up it might feel better.

- If the current one is common on Android then 👍 
- If not and it's an easy change, let's try 🙏 
- If it's complex and will take time, let's leave it for later 👍 


> Press the back button and verify you navigate to the Share screen (system bars should appear)

The system back button also gets hidden. The only way I see to dismiss this screen is by the ( x ) button (which I guess is fine).

On iOS we have the swipe down gesture as an alternative dismissal. It feels natural since the presentation is a modal which comes from the bottom up. And it's easier than reaching the high close button.

No need to implement this now, I'd rather move the button down closer to the thumb. So this is a wider discussion for later.

hamorillo:
Thanks for the review @etoledom!

About the status bar. I saw the iOS app and decided to keep it, so we are no longer hiding it. Also fixed the padding issue. 👍 

> The view presenting from the top left corner felt unexpected 🤔
> If the animation could come from straight up-down, or down-up it might feel better.

I've updated the animation so it appears from the top of the screen with a fade-in effect and disappears with the opposite animation. Let me know what you think.

> The system back button also gets hidden. The only way I see to dismiss this screen is by the ( x ) button (which I guess is fine).

I think this should be fine. Also, you should be able to show the navigation bar by swiping over the bottom area.

https://github.com/user-attachments/assets/d1726038-131e-4150-9c24-0484652bdd2e

etoledom (approved):
> I've updated the animation so it appears from the top of the screen with a fade-in effect and disappears with the opposite animation. Let me know what you think.

That's beautiful! 🙏 

> About the status bar. I saw the iOS app and decided to keep it, so we are no longer hiding it. Also fixed the padding issue. 👍

Perfect! Looks quite natural now 🎉 

> I think this should be fine. Also, you should be able to show the navigation bar by swiping over the bottom area.

Oh I didn't know that. Now I can corroborate that dismiss with back button works as described 👍 

---

Thank you for the improvements! 
🚢

PR #98, opened 2025-08-01
Delete User's Profile

hamorillo:
### Description

This PR implements the account deletion flow in the about dialog.

https://github.com/user-attachments/assets/cdbce984-7b70-498d-bad4-0466e64c8acc

There is no error handling, so if an error occurs during the flow, the dialog will be dismissed without providing more information.

**Note:** Design can be improved in the following PRs

### Testing Steps

- Launch the app
- Open the About dialog by pressing the three dots icon in the top right corner
- Select "Delete account"
- Verify the `Cancel` button works as expected
- Verify the `Delete account` button works as expected


etoledom (approved):
Looking good!

I have added a simple generic error alert dialog in case of error while trying to delete the account.

PR #99, opened 2025-08-04
Revert "[Tooling] Remove GitHub check about Prototype Build"

etoledom:
Reverts Automattic/Gravatar-Android#87

We need the prototype builds for testing the latest builds before release.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -91,10 +91,12 @@ steps:
     blocked_state: passed
 
   - label: 🏗️ Prototype Build
-    if: build.pull_request.id != null
     depends_on:
       - run-on-demand-prototype-build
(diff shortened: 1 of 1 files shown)


pinarol (approved):
Thank you 👍

iangmaia:
Hi @etoledom and @pinarol -- prototypes were normally available before #99.

#87 was removing purely the **cosmetic GitHub check**, which at request of @hamorillo I've removed based on [this discussion](https://github.com/Automattic/Gravatar-Android/pull/65#issuecomment-3109269033), but the **on-demand prototypes were still available from Buildkite**.
Part of this PR is actually important as it only allows for a PR prototype when you have a PR, otherwise the check will fail.

I've reverting the reversion 😅 here https://github.com/Automattic/Gravatar-Android/pull/101

PR #102, opened 2025-08-06
Share screen icons switch position

etoledom:
Closes GRA-676

### Description

This PR moves the icons on the share screen section headings to the leading side of the text.

CleanShot 2025-08-06 at 14 40 31@2x

### Testing Steps
- Open the app in the Share screen
  - Check the position of the icons in the section headings
- Tap on the Lock and the ( ! ) icon
  - Check that the modal appears.


AdamGrzybkowski (approved):
:shipit:

PR #103, opened 2025-08-06
Renaming make current and share buttons

etoledom:
Closes GRA-677

### Description

This PR makes these changes, renaming:

- `Make current` to `Select` in the avatar grid avatar action
- `Share` to `Share profile` in the headers app menu

CleanShot 2025-08-06 at 16 16 49@2x
CleanShot 2025-08-06 at 16 16 40@2x

### Testing Steps

- Run the app
- Check for both button title changes


homeUi/src/main/res/values/strings.xml
```diff
@@ -38,7 +38,7 @@
     <string name="gravatar_tab_avatar_upload_failure_image_too_big">The provided image exceeds the maximum size: 10MB</string>
     <string name="gravatar_tab_avatar_upload_error_action">Retry</string>
     <string name="gravatar_tab_avatar_upload_failure_dialog_remove_upload">Remove upload</string>
-    <string name="gravatar_tab_more_options_make_current">Make current</string>
+    <string name="gravatar_tab_more_options_make_current">Select</string>
```
AdamGrzybkowski: Because of the translation setup, it's always better to create a new String rather than modify an existing one. We already have the strings uploaded to GlotPress so we should follow that.

etoledom: Updated 👍

homeUi/src/main/res/values/strings.xml
```diff
@@ -52,7 +53,7 @@
     <string name="gravatar_tab_avatar_selection_failed">Unable to change your avatar. Try again.</string>
     <string name="gravatar_tab_avatar_deletion_failed">Couldn\'t delete the image. Try again.</string>
     <string name="gravatar_tab_topbar_menu_visit_profile">Visit your profile</string>
-    <string name="gravatar_tab_topbar_menu_share">Share</string>
+    <string name="gravatar_tab_topbar_menu_share">Share profile</string>
```
AdamGrzybkowski: We need to create a new one for this one as well.

etoledom: 🤦‍♂️ Done 👍

AdamGrzybkowski (approved):

PR #104, opened 2025-08-06
Add menu-selected avatar visual state

etoledom:
Closes GRA-679

### Description

This PR adds a visual mark when an avatar is presenting the actions button:

https://github.com/user-attachments/assets/28ddab08-feed-42b2-95b0-b5ea9427545d

### Testing Steps

- Run the app
- Go to the Avatars picker (Gravatar tab)
- Tap on an avatar to show the actions menu
  - Check that the avatar changes visually to signal selection.


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/SelectableAvatar.kt
```diff
@@ -97,6 +115,12 @@ private fun SelectableAvatar(
                     }
                 }
             }
+            .graphicsLayer(
+                alpha = animatedAlpha,
+                scaleX = animatedScale,
+                scaleY = animatedScale,
+                transformOrigin = TransformOrigin.Center
+            )
```
AdamGrzybkowski, line 114: 💡 Looks good, but I noticed that the ripple effect will now be bigger after you tap an avatar and expand beyond that scaled-down view.  You can fix this by moving the `.clickable {}` modifier to the bottom of the chain.

AdamGrzybkowski (approved):
Looks good, left one minor comment for consideration.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/gravatar/components/SelectableAvatar.kt
etoledom: Updated. 
I barely notice any ripple effect on tap. And I don't see any difference on this change (emulator and on device)

@AdamGrzybkowski - Could you take a look now? 👀 🙏

AdamGrzybkowski (approved):

PR #105, opened 2025-08-07
[Tooling] Add automation lane to fetch translations

iangmaia:
Fixes AINFRA-1033

### Description

This PR adds the lane `download_translations` that will fetch the translations from the Gravatar Android GlotPress projects ([login-ui](https://translate.wordpress.com/projects/gravatar/gravatar-android/login-ui/) and [home-ui](https://translate.wordpress.com/projects/gravatar/gravatar-android/home-ui/)) and add them to the project.

I've also ran the lane for the first time to initially add the files (though I noticed the only language with actual translations was Swedish).

### Testing Steps

1. Checkout this branch
2. Add translations to one of the GlotPress projects mentioned above
3. Run `bundle exec download_translations skip_commit:true`
4. Verify that the added translation has been correctly downloaded to the project.

### DIFF
--- a/homeUi/src/main/res/values/available_languages.xml
+++ b/homeUi/src/main/res/values/available_languages.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--Warning: Auto-generated file, do not edit.-->
+<resources xmlns:tools="http://schemas.android.com/tools">
+  <string-array name="available_languages" translatable="false" tools:ignore="InconsistentArrays">
+    <item>en_US</item>
+    <item>ar</item>
+    <item>de</item>
+    <item>es</item>
+    <item>fr</item>
+    <item>iw</item>
+    <item>in</item>
+    <item>it</item>
+    <item>ja</item>
+    <item>ko</item>
+    <item>nl</item>
+    <item>pt_BR</item>
+    <item>ru</item>
+    <item>sv</item>
+    <item>tr</item>
+    <item>zh_CN</item>
+    <item>zh_TW</item>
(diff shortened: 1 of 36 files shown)


homeUi/src/main/res/values/available_languages.xml
AliSoftware, line 1: (Nothing to change here, just me ranting)

That reminds me that this file (generated by our `gp_download_translations` helper from `release-toolkit`) is quite obsolete now…
 - This was a file used by the WordPress Android project only, so that they can [load it at runtime from their code to then have a custom language picker](https://github.com/wordpress-mobile/WordPress-Android/blob/8296f7562850b9ff477d785341b07dbfb336e56e/WordPress/src/main/java/org/wordpress/android/ui/prefs/language/LocalePickerViewModel.kt#L147)
 - But now WPAndroid have migrated to getting rid to the custom, in-app locale picker and using more modern Android APIs that offers system-level locale switching (similar to iOS supporting setting the locale per app in System Settings)… so I don't think it's used in WPAndroid anymore
 - And to my knowledge it's not used by any other project…

That shows the legacy of those helpers 😅

AliSoftware (approved):

AliSoftware:
Ooof we might want to update the Android linter settings to ignore `MissingTranslation` rule 😅

homeUi/src/main/res/values/available_languages.xml
iangmaia: Ahh thanks for the context 👍  I was curious why this file was being generated in the first place; just checked a few projects that still have it, and imagined it might still be used elsewhere 😅

PR #108, opened 2025-08-12
Fix avatar scaling issue when expanding the header in the Avatar tab

AdamGrzybkowski:
### Description

There's a problem with the avatar scaling animation when expanding the header after the image is selected. The avatar isn't scaling properly (only the outer component).

I noticed the issue would only happen when the image was not in the cache. Selecting an image that was already selected before would work just fine. The callback that sets the `oldImageKey` is invoked multiple times, so my guess is that it's related to extra recompositions... but not 100% sure. 

Adding a check and only setting the `oldImageKey` to a new value seems to solve the issue. 

Additionally, I've simplified the `GravatarAvatarWithShadow` code. The behavior should still be the same: a placeholder visible at all times, when loading a new avatar, the old one should stay visible until the new one is loaded. 

| Before | After |
|---|---|
| avatar_scale_before | avatar_scale_after |

### Testing Steps

1. Perform a fresh installation of the app (or delete the app cache)
2. Log in
3. Collapse the header
4. Select a new avatar
5. Expand the header 
6. Confirm the avatar looks good
7. Collapse the header
8. Select the previous avatar (to load from cache)
9. Expand the header 
6. Confirm the avatar looks good
7. Confirm that the old avatar is always visible untill the new one is loaded


etoledom (approved):
Nice! The issue is not reproducible any longer 🎉

PR #109, opened 2025-08-12
Introduce `crashlogging` / Sentry

wzieba:
Closes: AINFRA-954

### Description

This PR introduces integration of Sentry via our internal `crashlogging` wrapper. In addition, it introduces Sentry Gradle Plugin to send mapping files and source context.

#### Proguard deobfuscation + source context works fine

Screenshot 2025-08-12 at 17 39 02

#### Providing user data works fine as well

Screenshot 2025-08-12 at 18 04 30

### Testing Steps

Install the prototype build and observe Sentry dashboard for incoming events (there's a test error reported on app start).

### 🚨 Before merge 🚨

`git revert f672f18`


userComponent/build.gradle.kts
```diff
@@ -30,7 +30,7 @@ dependencies {
     implementation(libs.ktor.okhttp)
     implementation(libs.ktor.content.negotiation)
     implementation(libs.ktor.serialization.json)
-    implementation(libs.gravatar.core)
+    api(libs.gravatar.core)
```
wzieba, line 33: As `UserRepository#getProfile` exposes `com.gravatar.restapi.models.Profile`, Gravatar dependency has to be available for `userComponent` consumers.

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
```diff
@@ -76,4 +81,18 @@ class GravatarAndroidApplicationConventionPlugin : Plugin<Project> {
...
     }
+
+    private fun Project.configureSentry() {
+        configure<SentryPluginExtension> {
+            org.set("a8c")
+            projectName.set("gravatar-android")
+            authToken.set(System.getenv("SENTRY_AUTH_TOKEN"))
+
+            includeProguardMapping.set(System.getenv("CI").toBoolean())
+            includeSourceContext.set(System.getenv("CI").toBoolean())
```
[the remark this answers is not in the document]
wzieba: Nope, `toBoolean` extension supports nullable `String`

image

crashlogging/src/main/java/com/gravatar/crashlogging/ContextBasedLocaleProvider.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import android.app.Application
+import androidx.core.os.ConfigurationCompat
+import java.util.Locale
+
+internal class ContextBasedLocaleProvider(
+    private val context: Application,
+) : LocaleProvider {
+    override fun provideLocale(): Locale? {
+        return ConfigurationCompat.getLocales(context.resources.configuration)[0]
```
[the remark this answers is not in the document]
wzieba: While technically possible, this doesn't happen in other projects (e.g. Pocket Casts) so I wouldn't worry about it.

wzieba:
Requesting review from multiple people here to speed up the process (I start AFK the day after tomorrow). 1 review will be completely fine 👍 Thanks!

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.automattic.android.tracks.crashlogging.PerformanceMonitoringConfig
+import com.automattic.android.tracks.crashlogging.ReleaseName
+import com.gravatar.app.usercomponent.domain.repository.UserRepository
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.map
+
+internal class GravatarCrashLoggingDataProvider(
+    localeProvider: LocaleProvider,
+    userRepository: UserRepository
```
wzieba, line 15: I'm aware that this can circular DI dependency if `crashlogging` was about to be used in `userComponent` module. It's possible to address it one way or another, but until it won't be needed, I think there's no point to make this setup more complex now.

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
```diff
@@ -76,4 +81,18 @@ class GravatarAndroidApplicationConventionPlugin : Plugin<Project> {
             }
         }
     }
+
+    private fun Project.configureSentry() {
```
ParaskP7, line 85: Suggestion (💡): I saw and liked the `configureDetekt()` approach, why not doing the same, like creating a `Sentry` file and moving all such configuration within, removing it from this class and this way keeping it as small as possible:

`Sentry.kt`
```kotlin
package com.gravatar.app

import io.sentry.android.gradle.SentryPlugin
import io.sentry.android.gradle.extensions.SentryPluginExtension
import org.gradle.api.Project
import org.gradle.kotlin.dsl.apply
import org.gradle.kotlin.dsl.configure

internal fun Project.configureSentry() {
    apply<SentryPlugin>()

    configure<SentryPluginExtension> {
        org.set("a8c")
        projectName.set("gravatar-android")
        authToken.set(System.getenv("SENTRY_AUTH_TOKEN"))

        includeProguardMapping.set(System.getenv("CI").toBoolean())
        includeSourceContext.set(System.getenv("CI").toBoolean())

        autoInstallation.enabled.set(false)
        ignoredBuildTypes.set(listOf("debug"))
    }
}
```

---

Question (❓): Unrelated to the above, but while using the below command to test this setup:

`CI=true SENTRY_AUTH_TOKEN=<SENTRY_AUTH_TOKEN> ./gradlew assembleRelease`

I noticed that sometimes the mapping are uploaded:

`CLI`
image

`Sentry UI`
image

While other times they are not:

`CLI`
image

Any idea why? 🤔

userComponent/build.gradle.kts
ParaskP7: Suggestion (💡): Wouldn't it better if this remains as is `implementation(libs.gravatar.core)`, and instead, we add the `implementation(libs.gravatar.core)` dependency directly on the `crashlogging` module itself. This way we avoid having any module using that `userComponent` project to be exposed to that `implementation(libs.gravatar.core)` API by default, for no reason, just because we need it on the `crashlogging` module, wdyt? 🤔

`build.gradle.kts` (`crashlogging`)
```kotlin
dependencies {
    implementation(project(":userComponent"))
    implementation(libs.gravatar.core)

    ...
}
```

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override val buildType = BuildConfig.BUILD_TYPE
+
+    override val enableCrashLoggingLogs = BuildConfig.DEBUG
+
+    override val locale = localeProvider.provideLocale()
+
+    override val performanceMonitoringConfig = PerformanceMonitoringConfig.Disabled
+
+    override val releaseName = ReleaseName.SetByTracksLibrary
```
ParaskP7: Suggestion (💡): On other projects of ours, we have a configuration such as the below, maybe worth adding it here as well:

```kotlin
    override val releaseName = if (BuildConfig.DEBUG) {
        ReleaseName.SetByApplication(DEBUG_RELEASE_NAME)
    } else {
        ReleaseName.SetByTracksLibrary
    }

    companion object {
        const val DEBUG_RELEASE_NAME = "debug"
    }
```

Not sure if that was considered already, but wdyt? 🤔

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    override val sentryDSN: String = BuildConfig.SENTRY_DSN
+
+    override val user: Flow<CrashLoggingUser> = userRepository.getProfile().map { profile ->
+        CrashLoggingUser(
+            userID = profile?.userId?.toString(),
+            email = null,
```
ParaskP7: Suggestion (💡): What if we use `profile?.hash` as email, just to have that extra info when and if needed, instead of no email whatsoever? 🤔

FYI: See `Profile.kt` model class below:

```kotlin
@JsonClass(generateAdapter = true)
public class Profile internal constructor(
    // The SHA256 hash of the user's primary email address.
    @Json(name = "hash")
    public val hash: kotlin.String,
    ...
}
```

image

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
```diff
@@ -0,0 +1,55 @@
...
+            username = profile?.displayName,
+        )
+    }
+
+    override fun crashLoggingEnabled(): Boolean = true
```
ParaskP7: (Super) Minor (🔍): Consider removing explicit type specification from everywhere possible:

```kotlin
internal class GravatarCrashLoggingDataProvider(
    localeProvider: LocaleProvider,
    userRepository: UserRepository
) : CrashLoggingDataProvider {

    ...

    override val user = userRepository.getProfile().map { profile ->
        CrashLoggingUser(
            userID = profile?.userId?.toString(),
            email = null,
            username = profile?.displayName,
        )
    }

    override fun crashLoggingEnabled() = true

    override fun extraKnownKeys() = emptyList<ExtraKnownKey>()

    override fun provideExtrasForEvent(
        currentExtras: Map<ExtraKnownKey, String>,
        eventLevel: EventLevel
    ) = emptyMap<ExtraKnownKey, String>()

    override fun shouldDropWrappingException(
        module: String,
        type: String,
        value: String
    ) = false
}
```

ParaskP7 (approved):
👋 @wzieba !

I have reviewed and tested this PR as per the instructions, everything works as expected, nicely done introducing Sentry here! 💯 x 💯 ^ 💯

---

I've tested the overall solution building the release build on CLI using this command below, and everything is working exactly as expected (give-or-take the mappings upload, see [question](https://github.com/Automattic/Gravatar-Android/pull/109#discussion_r2272791769)).

`CI=true SENTRY_AUTH_TOKEN=<SENTRY_AUTH_TOKEN> ./gradlew assembleRelease`

ProGuard Deobfuscation + Source Context: ✅ 

image

User Data: ✅ 

image

Sentry ProGuard Mapping Upload: ✅ (ish, , see [question](https://github.com/Automattic/Gravatar-Android/pull/109#discussion_r2272791769))

image

image

-----

I have left one question (❓), few suggestions (💡) and one minor (🔍) comments for you to consider. I am going to approve this PR anyway, since none is blocking. I am **NOT** going to merge this PR yet to give you some time to apply any of my suggestions. However, feel free to ignore them and merge the PR yourself.

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
AdamGrzybkowski: Although not ideal and I would imagine pushing the new profile (or rather `CrashLoggingUser`) when updated outside of the module rather than depending on the `:userComponent` I think this is okay for now. 

>if crashlogging was about to be used in userComponent module.

I assume that fatal crashes are reported automatically. So that would only be needed if we wanted to report non-fatal ones, correct?

userComponent/build.gradle.kts
AdamGrzybkowski: Modules that depend on `:userComponent` will likely need the `com.gravatar.restapi.models.Profile` type so I think it's fine to use `api`.

AdamGrzybkowski (approved):
I can see the crashes ✅  I left one question but other than that it looks good.

userComponent/build.gradle.kts
ParaskP7: Oh cool, if that's the case then 💯 this is the correct approach, thanks for chipping-in @AdamGrzybkowski ! 🙇

PS: I am resolving this. ✅

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
wzieba: > I saw and liked the `configureDetekt()` approach, why not doing the same

`configureDetekt` is used in multiple plugins, `configureSentry` is used only in once. I think it's better for readability to have things `private` when possible.

> While other times they are not:

I guess you didn't change the code every time - Sentry sees that the corresponding mapping is already uploaded, and doesn't reupload.

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
wzieba: > Although not ideal and I would imagine pushing the new profile 

I get you, but I also like the reactiveness of this solution - it'd be great if code related to `user` wouldn't have to remember about pushing a new profile to things like "crash logging".

> I assume that fatal crashes are reported automatically. So that would only be needed if we wanted to report non-fatal ones, correct?

True, you're correct! 👍

build-logic/convention/src/main/kotlin/GravatarAndroidApplicationConventionPlugin.kt
ParaskP7: > configureDetekt is used in multiple plugins, configureSentry is used only in once.

👍 @wzieba  💯 

> I think it's better for readability to have things private when possible.

Up to you, I would still have that as a separate `Sentry` file, for that exact purpose, better readability. Maybe, with `private` you don't mean better readability but actually closer access so that this is not used on a more wider scope, I get that... 👍

> I guess you didn't change the code every time - Sentry sees that the corresponding mapping is already uploaded, and doesn't reupload.

Ah, nice, that's what I was assuming at the end, but good to have that confirmed, thanks! 💯

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
ParaskP7: Done: 0399e57cf9b522c7e5cad3d564048f8b9c2be41a

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
ParaskP7: Done: c5383e7ef0759b2fe7faccd7f1f42a6c31b850ee

crashlogging/src/main/java/com/gravatar/crashlogging/GravatarCrashLoggingDataProvider.kt
ParaskP7: Done: a0f1392a6c9038b61349dd79381a73b905fdc3cf

ParaskP7:
Thanks for the update @wzieba , another ✅ from me! :shipit:

PR #110, opened 2025-08-13
Fix header overlay darkness on API 11 and lower 

AdamGrzybkowski:
### Description

The `.blur` Modifier we use to blur the header image works only on API 12 and higher. We have an additional overlay with `Color.Black(alpha = 0.15f)` but that wasn't enough without the blur. The first commit https://github.com/Automattic/Gravatar-Android/commit/ba92a807afe0fdcb94df6c95a16c28a1e8d34fc7 changes the alpha for Android 11 and lower.

Why not blur on older APIs? I've left a comment about this [here](https://linear.app/a8c/issue/GRA-487/blur-on-headers-is-not-working-as-expected-in-api-28#comment-c5ebf8ef). We should be able to add this, but the ProfileHeader needs some adjustments to make it look nice. 

| Before | After |
|---|---|
| overlay_API_28_before | overlay_API_28_after |

This change made it more obvious that the status bar icons are not properly tinted. https://github.com/Automattic/Gravatar-Android/commit/019a7969942c851c4cb8149440509588ae6222e9 commit fixes that by setting a proper status bar appearance. I create a new `Screen` component that handles this logic, so each screen component should use this one as a wrapper now. This will come in handy for the Screen analytics events as well. 

| Before | After |
|---|---|
| status_bar_before | light_status_bar_after |

Additionally, I've noticed that forcing a dark theme on the Profile header forces the dark theme on the `TopBarPickerPopup` component. This was fixed here as well.

| Before | After |
|---|---|
| popup_menu_before | popup_menu_after |

### Testing Steps

1. Launch the app on API 11 or lower and test the header overlay.
2. Switch the device to light mode and open the top menu in the app - confirm it uses light mode as well. 
3. Verify that the status bar icons are always white in the headers (Profile, Avatar and Share tabs), regardless of the device's dark/light mode. `LoginScreen` should set the appearance based on the device's color mode, the icons should change the color.

### DIFF
--- a/design/build.gradle.kts
+++ b/design/build.gradle.kts
@@ -9,6 +9,7 @@ android {
 
 dependencies {
     implementation(libs.androidx.material3)
+    implementation(libs.androidx.core.ktx)
 
     testImplementation(libs.junit)
     testImplementation(project(":testUtils"))
--- a/design/src/main/kotlin/com/gravatar/app/design/components/Screen.kt
+++ b/design/src/main/kotlin/com/gravatar/app/design/components/Screen.kt
@@ -1,0 +1,21 @@
+package com.gravatar.app.design.components
+
+import androidx.compose.foundation.isSystemInDarkTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.SideEffect
+import androidx.compose.ui.platform.LocalView
+import androidx.core.view.WindowCompat
+
+@Composable
+fun Screen(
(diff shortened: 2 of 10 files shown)


etoledom (commented):
Looking great!

I've seen the new light status bar and the correct color of the menu on the main menu.

I couldn't test the darkened header in the older API. 
@pinarol - could you take a fast look at this one? I've run the on-demand build.

pinarol:
I observe the same thing with my Android 9 device 🎉

etoledom (approved):

PR #111, opened 2025-08-13
Don't show the NoAvatarSelected banner when the user is uploading the first avatar

AdamGrzybkowski:
### Description

This PR fixes a small issue where we show the `NoAvatarSelected` banner while uploading the first avatar. 

I've added a screenshot test to verify this scenario. 

### Testing Steps

1. Launch the app
2. Remove all your avatars
3. Upload a new one
4. Confirm the banner is not shown while uploading the avatar


etoledom (approved):
Good catch!
This solves the problem 🎉

PR #112, opened 2025-08-18
Add verified accounts to the Share tab

AdamGrzybkowski:
### Description

This PR lists all the verified accounts of the user in the Share tab public section. By default, each account is shared.

**Profile's Verified Accounts storage** (commit https://github.com/Automattic/Gravatar-Android/commit/9d75cc7fcfd678407b87f5e172a1482347224cc8):
A new table and entity were created for `VeriefiedAccount` model. 

**Storing user Share preferences for each account** (commit https://github.com/Automattic/Gravatar-Android/commit/d04edae3e46b1d100699d7a0f8809d55a125eb14):
User share preferences for the verified accounts are stored under one shared key in the format Json format. Using the URL as a key is taken from the iOS app. Why a shared, single key? An alternative could be to use something like  `key_account_url` for each account, but with such an approach, we would always have to send all the accounts to get the settings. With a single key, we can get whatever is saved in the prefs easily.

Screenshot 2025-08-18 at 12 56 40

### Testing Steps

1. Make sure you have some verified accounts attached to your account
2. Launch the app and open the Share tab
3. Scroll down and confirm that the verified accounts are visible and are checked 
4. Verify the vCard data includes them (this will depend on your phone, but for me on the Pixel the extra URLs are ignored when I scan the QR code, but are used when I download the .vcard file and open it on the device)
5. Disabled one of the accounts
6. Verify the vCard again
7. Kill and open the app
8. Verify the state in the share tab is restored

DB migration:
1. Install the app from trunk
2. Log in
3. Update the app installing the app from this PR 
4. Confirm the DB was properly migrated (no crashes, the data was still there)


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/model/VCard.kt
```diff
@@ -65,6 +70,12 @@ internal class VCard private constructor(
     // We've seen issues with newlines in the vCard content causing problems when importing the contact so removing them
     private fun String.escaped() = this.replace("\n", " ")
 
+    private fun List<URL>.toVCardFormat(): String {
+        return joinToString(separator = "\n") { url ->
+            "URL;TYPE=\"${url.label}\":${url.url}"
+        }
+    }
```
AdamGrzybkowski, line 77: @etoledom I don't think you need to review the full code, but it would be great if you could verify the vCard data (and whether it matches with iOS).

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/model/VCard.kt
```diff
@@ -65,6 +70,12 @@ internal class VCard private constructor(
     // We've seen issues with newlines in the vCard content causing problems when importing the contact so removing them
     private fun String.escaped() = this.replace("\n", " ")
 
+    private fun List<URL>.toVCardFormat(): String {
+        return joinToString(separator = "\n") { url ->
```
hamorillo, line 74: ⚠️  I haven't reviewed all the code yet, but I get an error when trying to import the exported vCard in my Pixel due to a missing `\n` after the last URL.

hamorillo (changes requested):
Apart from [this comment](https://github.com/Automattic/Gravatar-Android/pull/112#discussion_r2282758569), code looks good and works as expected. Tested the DB migration and vCard generation.

⛏️ Totally nitpicking, I think this part from the PR Description no longer applies as you've refactored it to JSON serialization, which in my opinion, looks better :)

> User share preferences for the verified accounts are stored under one shared key in the format [url=isChecked;url2=isChecked;...].

💡 This is not related to this PR, but I wonder if we should allow users to at least disconnect verified accounts from the profile section. Or include some kind of banner/link to let them know how to manage them. cc: @pinarol @etoledom 
Also, we are ignoring the `isHidden` which can create some misleadings.

AdamGrzybkowski:
> ⛏️ Totally nitpicking, I think this part from the PR Description no longer applies as you've refactored it to JSON serialization, which in my opinion, looks better :)

 🤦 yes... fixed.


>Also, we are ignoring the isHidden which can create some misleadings.

 I was thinking about it. We could use it to default to false at least.

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/share/model/VCard.kt
AdamGrzybkowski: ...correct, I have removed this by accident when refactoring some code 🤦 Should be fixed now.

hamorillo (approved):
🚀 

>  I was thinking about it. We could use it to default to false at least.

Sounds like a good option. 🙂

PR #113, opened 2025-08-18
Rename tab: Share -> QR

AdamGrzybkowski:
### Description

Tab rename: Share -> QR

Screenshot 2025-08-18 at 15 58 55

### Testing Steps

1. Launch the app
2. Verify the tab name


hamorillo (commented):
Code changes look good!

⚠️ Screenshot tests are passing, probably due to the `changeThreshold`, but we need to regenerate the screenshots.

hamorillo:
Thinking about the threshold, an idea, if possible, could be to apply the threshold only to the "problematic" test, for example, using an annotation. Not related to this PR, just an idea to discuss.

AdamGrzybkowski:
You're right. I've tried changing the threshold but other tests started to fail, but not the ones with the updated copy...

hamorillo (approved):
🚀

PR #114, opened 2025-08-19
Privacy Settings

AdamGrzybkowski:
### Description

This PR adds the Privacy Setting bottom sheet with an option to opt-out of analytics and/or Crashlytics.

The `PrivacySettingsStorage` stores the information locally and will be cleared during logout. 

Screenshot_20250819_121256

⚠️ Revert https://github.com/Automattic/Gravatar-Android/pull/114/commits/8720dc2cf491a06caccb244db133c24eaa859958 before merging 
### Testing Steps

1. Launch the app
2. Log in
3. Open the AboutAppDialog via `...` menu button
4. Tap Privacy Settings
5. Confirm the Privacy Settings are displayed 
6. Change the settings values
7. Kill and open the app
8. Open the Privacy Settings again and verify the values are restored

There's an extra commit https://github.com/Automattic/Gravatar-Android/pull/114/commits/8720dc2cf491a06caccb244db133c24eaa859958 to test crash reporting.

1. Install the prototype build
1. Enable crash reporting
2. Tap `Done` button in the About app dialog
3. Confirm the error is logged in Sentry
4. Disable crash reporting 
5. Tap `Done` button in the About app dialog
6. Confirm the error doesn't appear in Sentry


homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/topbar/components/about/AboutAppDialog.kt
```diff
@@ -161,9 +166,9 @@ internal fun AboutAppDialogContent(
                 }
             )
             DialogText(
-                text = stringResource(R.string.about_app_dialog_privacy_policy),
+                text = stringResource(R.string.about_app_dialog_privacy_settings),
                 modifier = Modifier.clickable {
-                    context.openPrivacyPolicy()
+                    onPrivacySettingsClicked()
```
AdamGrzybkowski, line 171: ❓ Should we invoke onDismiss as well here?

hamorillo: We are not dismissing it for the delete flow so that I would say no. (Or, yes, if we apply the same approach to both flows).

homeUi/src/main/kotlin/com/gravatar/app/homeUi/presentation/home/components/privacySetting/PrivacySettingsBottomSheet.kt
```diff
@@ -0,0 +1,224 @@
...
+}
+
+private fun Context.openPrivacyPolicy() {
+    val intent = Intent(Intent.ACTION_VIEW, PRIVACY_POLICY_URL.toUri())
+    startActivity(intent)
+}
+
+@Composable
+private fun PrivacySettingsCard(
+    settingTitle: String,
```
hamorillo, line 170: ⛏️ 
```suggestion
    @StringRes settingTitle: Int,
    @DrawableRes settingIcon: Int,
    @StringRes settingDescription: Int,
```

hamorillo (approved):
Looks good! 

❓Probably in a different PR, but should we use those privacy settings to disable the `CrashLogging` somehow? I would say we should also do the same in `TracksTracker`.

hamorillo (approved):
🚀

PR #116, opened 2025-08-19
Add `screen_view` and `screen_leave` analytic events

hamorillo:
### Description

This PR adds the properties to the Event class in the analytics module. Also, to verify that it works, I'm adding the `screen_view` and `screen_leave` events on the `Screen` composable.

This way, we can also work in parallel to add analytics, avoiding some conflicts.

### Testing Steps

- Verify you can see the events on `tracks` (you can use the filter `gravatarandroid%`)
- You can also verify that the event is tracked by adding a `flush` after each event and analyzing the network requests.

### DIFF
--- a/analytics/src/main/kotlin/com/gravatar/analytics/Event.kt
+++ b/analytics/src/main/kotlin/com/gravatar/analytics/Event.kt
@@ -4,10 +4,14 @@ import org.json.JSONObject
 
 interface Event {
     val name: String
-    val properties: Properties?
-        get() = null
+    val properties: Map<String, Any>
+        get() = emptyMap()
+}
 
-    interface Properties {
-        fun toJson(): JSONObject
+internal fun Map<String, Any>.asJson(): JSONObject {
+    return JSONObject().apply {
+        this@asJson.forEach { (key, value) ->
+            put(key, value)
+        }
     }
 }
--- a/analytics/src/main/kotlin/com/gravatar/analytics/Event.kt
+++ b/analytics/src/main/kotlin/com/gravatar/analytics/Event.kt
@@ -1,5 +1,13 @@
 package com.gravatar.analytics
 
+import org.json.JSONObject
+
 interface Event {
     val name: String
+    val properties: Properties?
+        get() = null
+
+    interface Properties {
+        fun toJson(): JSONObject
+    }
 }
--- a/analytics/src/main/kotlin/com/gravatar/analytics/tracks/TracksTracker.kt
+++ b/analytics/src/main/kotlin/com/gravatar/analytics/tracks/TracksTracker.kt
@@ -3,6 +3,7 @@ package com.gravatar.analytics.tracks
 import com.automattic.android.tracks.TracksClient
 import com.gravatar.analytics.Event
 import com.gravatar.analytics.Tracker
+import com.gravatar.analytics.asJson
 import java.util.UUID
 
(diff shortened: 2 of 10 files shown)


analytics/src/main/kotlin/com/gravatar/analytics/tracks/TracksTracker.kt
```diff
@@ -6,14 +6,20 @@ import com.gravatar.analytics.Tracker
 import java.util.UUID
 
 internal class TracksTracker(private val tracksClient: TracksClient) : Tracker() {
+
+    internal companion object {
+        const val TRACKS_EVENT_NAME_PREFIX = "gravatarandroid_"
```
AdamGrzybkowski, line 12: ❓ Is that a Track's requirement to prefix all the events?y

analytics/src/main/kotlin/com/gravatar/analytics/Event.kt
```diff
@@ -1,5 +1,13 @@
...
+import org.json.JSONObject
+
 interface Event {
     val name: String
+    val properties: Properties?
+        get() = null
+
+    interface Properties {
+        fun toJson(): JSONObject
+    }
```
AdamGrzybkowski: 💡 What do you think about replacing this with:
```kotlin
interface Event {
    val name: String
    val properties: Map<String, Any>
        get() = emptyMap()

}

internal fun Map<String, Any>.asJson(): JSONObject {
    return JSONObject().apply {
        this@asJson.forEach { (key, value) ->
            put(key, value)
        }
    }
}

```

This way, we don't have to create a new class for each event that has properties. The `DesignEvent` class would look like this:

```kotlin
sealed class DesignEvent : Event {

    data class ScreenView(val screen: String) : DesignEvent() {
        override val name: String = "screen_view"
        override val properties: Map<String, Any> = mapOf(
            "screen" to screen,
        )
    }

    data class ScreenLeave(val screen: String) : DesignEvent() {
        override val name: String = "screen_leave"
        override val properties: Map<String, Any> = mapOf(
            "screen" to screen,
        )
    }
}
```

design/src/main/kotlin/com/gravatar/app/design/components/Screen.kt
```diff
@@ -2,20 +2,40 @@ package com.gravatar.app.design.components
...
     SideEffect {
         val window = (view.context as? android.app.Activity)?.window
         val controller = window?.let { WindowCompat.getInsetsController(it, view) }
         controller?.isAppearanceLightStatusBars = appearanceLightStatusBars
     }
+
+    // Track when the screen becomes visible
+    LaunchedEffect(screenName) {
+        tracker.trackEvent(DesignEvent.ScreenView(screenName))
+    }
```
AdamGrzybkowski: ❓ Can't this be done in `DisposableEffect`?

analytics/src/main/kotlin/com/gravatar/analytics/tracks/TracksTracker.kt
hamorillo: As far as I remember, yes. In fact, we should use that concrete prefix as it was added to the allow list.

design/src/main/kotlin/com/gravatar/app/design/components/Screen.kt
hamorillo: Done!

AdamGrzybkowski (approved):
:shipit:

PR #117, opened 2025-08-20
Use Android App link for redirect URI

hamorillo:
### Description

This PR migrates from the custom schema used for the OAuth redirection to use [Android App Link](https://developer.android.com/training/app-links/verify-android-applinks).

The behaviour in the login flow should be identical from the users' point of view. 

All the backend configuration is already done, so it should work smoothly.

**Note:** Don't forget to update your project secrets.

### Testing Steps

❗ Test both debug and prototype build as we need to verify that both fingerprints are correct in the backend.

- Verify that the app works as expected after the login (OAuth) flow. Sign out and sign in.
- [Use Prototype Build] Verify that the app works as expected after the login (OAuth) flow. Sign out and sign in.


AdamGrzybkowski (approved):
Both build types worked :shipit:

PR #118, opened 2025-08-20
Bind privacy settings and userInfo with the TracksTracker

AdamGrzybkowski:
### Description

This PR binds the PrivacySettings and the userId with the `TracksTracker`. 

To not create a dependency on `:userComponent` in the `:analytics` module, the `:analytics` module exposes a public interface `TrackerSetupDataProvider` that is implemented and provided in the `:app` module. 

`TracksTracker` uses that provider to reactively update the state.

### Testing Steps
1. Enable tracking
2. Go through the app 
3. Confirm the events are sent (in the Tracks UI or via network requests)
4. Disable the tracking
5. Go through the app 
3. Confirm the events aren't sent (in the Tracks UI or via network requests)


analytics/src/main/kotlin/com/gravatar/analytics/Tracker.kt
```diff
@@ -1,7 +1,6 @@
 package com.gravatar.analytics
 
 abstract class Tracker {
-    abstract var userId: String?
```
AdamGrzybkowski, line 4: I couldn't find a reason why we would need that here now so I removed it.

analytics/src/main/kotlin/com/gravatar/analytics/TrackerSetupData.kt
```diff
@@ -0,0 +1,11 @@
+package com.gravatar.analytics
+
+data class TrackerSetupData(
```
AdamGrzybkowski, line 3: Not sure about the name. Any ideas?

app/src/main/java/com/gravatar/app/analytics/AppEvent.kt
AdamGrzybkowski, line 1: This wasn't used anymore.

app/src/test/kotlin/com/gravatar/app/AppTrackerSetupDataProviderTest.kt
```diff
@@ -0,0 +1,149 @@
...
+        privacySettingsFlow = MutableSharedFlow()
+        profileFlow = MutableSharedFlow()
+        provider = AppTrackerSetupDataProvider(
+            getPrivacySettings = getPrivacySettings,
+            userRepository = userRepository,
+        )
+    }
+
+    @Test
+    fun `emits ENABLED when analytics enabled with non-null user id`() = runTest {
```
hamorillo: ⛏️ **Nitpicking:**  I think the test name can lead to some confusion, as it looks like analytics will be only enabled with a non-null user ID when there is no relation between them.

analytics/src/main/kotlin/com/gravatar/analytics/TrackerSetupData.kt
hamorillo: What about `TrackingContext` or `TrackerConfig`?

hamorillo (approved):
🚀  Looks great!

analytics/src/main/kotlin/com/gravatar/analytics/TrackerSetupData.kt
AdamGrzybkowski: hmm not sure - do you have a strong preference for any of those?

app/src/main/java/com/gravatar/app/AppTrackerSetupDataProvider.kt
```diff
@@ -0,0 +1,31 @@
...
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+
+class AppTrackerSetupDataProvider(
+    private val getPrivacySettings: GetPrivacySettings,
+    private val userRepository: UserRepository,
+) : TrackerSetupDataProvider {
+    override fun getTrackerSetupData(): Flow<TrackerSetupData> {
+        val userIdFlow: Flow<String?> = userRepository.getProfile().map { it?.userId?.toString() }.distinctUntilChanged()
```
hamorillo: 🛑 We need to use the WPCOM userName for tracks instead of the userId.

```suggestion
        val userIdFlow: Flow<String?> = userRepository.getProfile().map { it?.userLogin }.distinctUntilChanged()
```

That suggestion is not enough, as we need to persist it in the DB.

analytics/src/main/kotlin/com/gravatar/analytics/TrackerSetupData.kt
hamorillo: I like `TrackingContext`, but I don't have a strong opinion. Feel free to keep `TrackerSetupData` if you want.

